PR #4, opened 2022-08-26
Porting #SF patch #565 Real Time Clock /CMOS fix

stlintel:
by Michele Giacomone

Detailed description:

  -Observed issues

   Due to some limitations only dates between 1980 and 2038 can be
   used in a reliable way.
   Also, bochs incorrectly assumes a linear correspondence between
   the data returned by the <time.h> functions localtime() and
   mktime(), and isn't setting the latter properly.
   Bochs keeps its internal time value dependent to these functions
   after setup, assuming that their internal settings won't change
   on the go - which is not the case.
   In my OS, and in my timezone, this leads to incorrect startup values
   for 5 months each year and unreliable values if the simulation is
   kept going for a long time. (a feedback between localtime() and
   mktime() is created which keeps shifting back the time)
   Also, the RTC simulation is not realistic since the clock fixes
   itself across DST changes, without updating any DST related flag,
   a behavior that no guest OS expects.

  -Proposed fix

   This is implemented in such way that no bochs' previous behavior
   is changed, a part from the broken ones, with legacy in mind
   == the user can keep using bochs exactly as before knowing nothing
      of this patch

   +Make the internal s.timeval variable a Bit64s, so it can fit all
    values that the cmos can correctly represent, reported below:
    MIN     setting  -62167219200 =>  0000/01/01 SAT  0:00:00
    MAX BCD setting  253402300799 =>  9999/12/31 FRI 23:59:59
    MAX BIN setting  745690751999 => 25599/12/31 FRI 23:59:59
    And then fix each reference to these so it can handle such values
    And make bochs correctly wrap around for under/overflows, so that
    only the most significant bits of the century are lost.

   +Do the same thing to the bochs time0 parameter, so all the above
    values can be chosen at startup (despite being now legal values,
    1 and 2 will still be treated as "local" and "utc"). Note that
    normally only BCD settings are valid since bochs' CMOS defaults
    to such operating mode - the only way to use the binary range
    is by loading a cmos memory map.

   +Make the internal s.timeval variable independent from external
    factors. This means providing a small set of time handling
    functions, contained in "iodev/utctime.h", which must work in
    any environment in which bochs compiles, accessing no external
    resource. This also means that after startup, s.timeval will only
    be changed internally, and no call to the OS time functions will
    be made.

   +Make the internal s.timeval variable timezone independent, to
    have a linear correlation between its values and valid CMOS
    settings. To make it easier, s.timeval is gonna be treated as
    if the current timezone was UTC: so,
     - if the user selects UTC as time0, s.timeval will become current
       time(NULL)
     - if the user selects localtime, s.timeval will be computed as
       the value which will display the same broken down time as
       localtime(&now)
     - if the user inputs a time formatted string the proper s.timeval
       to displayed will be easily calculated,
     - if the user inputs a starting time value, s.timeval will be
       computed as the value which will display the same broken down
       time as localtime(&user_input) to ensure the same operation as
       before.
    A "tz=utc" is displayed when bochs prints out the current time
    value, to warn users about the difference in meaning between the
    internally kept time value and the value they can set through
    the "time0=" parameter. This might be changed to communicate
    instead the time value they can input to get the same setting,
    but performing such calculation (except for the startup time)
    suffers from all the mktime()/localtime() problems listed above
    so I did not do it.
    The range of "time0" is automatically adjusted so all users in
    all time zones can set any legal value despite "time0=" having a
    local meaning.

  A thorough explanation of what I did and why can be found in the
  "iodev/utctime.h" library header.


stlintel:
I hope someone who have capability / expertise of compiling Bochs BIOS could help testing this PR.
I trust the author and believe this has added value to the project but still somehow need to validate it.

bochs/cpu/sha.cc
```diff
@@ -256,10 +256,10 @@ void BX_CPP_AttrRegparmN(1) BX_CPU_C::SHA1RNDS4_VdqWdqIbR(bxInstruction_c *i)
     E[n+1] = D[n];
   }
 
-  op1.xmm32u(0) = A[4];
```
Nable80, line 259: This change is definitely unrelated to this PR

bochs/gui/textconfig.cc
```diff
@@ -202,12 +202,12 @@ int ask_int(const char *prompt, const char *help, Bit32s min, Bit32s max, Bit32s
     }
     if ((clean[0] == '?') && (strlen(help) > 0)) {
       bx_printf("\n%s\n", help);
-      bx_printf("Your choice must be an integer between %u and %u.\n\n", min, max);
+      bx_printf("Your choice must be an integer between %ld and %ld.\n\n", min, max);
```
Nable80, line 205: ``long`` is a 32-bit type on some platforms, ``%l_`` is definitely not enough for 64-bit values.

bochs/cpu/sha.cc
stlintel: Yes, it is already in master

Nable80: ``git rebase`` often comes to the rescue when something has already landed into the master

vruppert:
I'm interested in testing this patch, but I'm unable to resolve the issues above. The conflicting file is not a part of the original patch.

stlintel:
> I'm interested in testing this patch, but I'm unable to resolve the issues above. The conflicting file is not a part of the original patch.

Resolved

bochs/gui/textconfig.cc
vruppert: Using the portable symbol FMT_LL (from osdep.h) should fix this issue.

vruppert:
From my point of view these changes are ready to review and test. I guess the original author never compiled a BIOS, but now this part of the patch also works as expected.

vruppert (approved):

PR #10, opened 2022-08-29
add layout commands for dbg

therealdreg:
Layout support (our GDB-TUI for poors :D):

```
addlyt file - cause debugger to execute a script file every time execution stops.
    Example of use: 1. Create a script file (script.txt) with the following content:
             regs            
             print-stack 7
             u /10
             <EMPTY NEW LINE>
    2. Execute: addlyt "script.txt"
    Then, when you execute a step/DebugBreak... you will see: registers, stack and disasm.
```
```
remlyt - stops debugger to execute the script file added previously with addlyt command.
```
```
lyt - cause debugger to execute script file added previously with addlyt command.
    Use it as a refresh/context.
```

Example of use:

layout

example_of_use_layout

Re-execute layout:

reload_layout

IMO, this will be **one of the most used features**.  (search_in_history or write same commands each time its a pain in the ass)

One idea: the user can have some script files with different layouts/things and change between them using **addlyt** command

To keep this feature short and simple I am using **bx_nest_infile()** (used by **source** command).

Only 1 layout/file is supported at time.

### DIFF
--- a/bochs/bx_debug/dbg_main.cc
+++ b/bochs/bx_debug/dbg_main.cc
@@ -62,6 +62,8 @@ static char tmp_buf_prev[512];
 static char *tmp_buf_ptr;
 static char *argv0 = NULL;
 
+static char layoutpath[BX_MAX_PATH];
+
 static FILE *debugger_log = NULL;
 
 static struct bx_debugger_state {
@@ -2258,6 +2260,29 @@ void bx_dbg_disassemble_current(int which_cpu, int print_time)
   }
 }
 
+void bx_dbg_addlyt(const char* new_layoutpath)
+{
+  if (strlen(new_layoutpath) > sizeof(layoutpath) - 1) {
+    dbg_printf("Error: %s path too long!\n", new_layoutpath);
+    return;
+  }
+  strcpy(layoutpath, new_layoutpath);
+}
+
+void bx_dbg_remlyt(void)
+{
+  layoutpath[0] = '\0';
+}
+
+void bx_dbg_lyt(void)
+{
+  if ('\0' == layoutpath[0]) {
+    return;
+  }
+
+  bx_nest_infile(layoutpath);
+}
+
 void bx_dbg_print_guard_results(void)
 {
   unsigned cpu, i;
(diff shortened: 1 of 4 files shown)


bochs/bx_debug/dbg_main.cc
```diff
@@ -2258,6 +2260,26 @@ void bx_dbg_disassemble_current(int which_cpu, int print_time)
   }
 }
 
+void bx_dbg_addlyt(const char* new_layoutpath)
```
stlintel, line 2263: doesn't feel safe to me.
if new_path is too long - need to have a nice error message and not copy part of the path and hope for good

bochs/bx_debug/dbg_main.cc
```diff
@@ -2258,6 +2260,26 @@ void bx_dbg_disassemble_current(int which_cpu, int print_time)
...
+}
+
+void bx_dbg_lyt(void)
+{
+  if ('\0' == layoutpath[0]) {
+    return;
+  }
+
+  bx_nest_infile(layoutpath);
+}
```
stlintel, line 2281: What will cause the debugger to run the script ?

stlintel (changes requested):

bochs/bx_debug/dbg_main.cc
therealdreg: This :D 

I am emulating a source command haha. Its tricky but simple and it works fine. 

You can try it! just clone my master: 

git clone https://github.com/therealdreg/Bochs.git

bochs/bx_debug/dbg_main.cc
therealdreg: oki! solved :D

bochs/bx_debug/dbg_main.cc
therealdreg: image

therealdreg: I think something weird happened with the previous commit, now it's fixed :D

therealdreg:
Any suggestion? for me its ok

stlintel (approved):

PR #28, opened 2023-02-16
patched an issue with the HPET

ImWuX:
Patched an issue with the HPET where writing an u64 with HPET_TN_SETVAL set, would only write the lower 32bits into the comparator. This is because the write is getting split into two 32bit writes and the first 32bit write will disable HPET_TN_SETVAL.


stlintel (approved):

stlintel:
Approved and merged !

PR #29, opened 2023-02-17
USB emulation enhancement and fixes

fysnet:
Minor:
  1) Check speed indicator in the Transfer Descriptor matching the device speed.
  2) Max packet size check, checking that the amount requested/sent is within limits.
  3) Checking that the first packet sent is the GetDescriptor Request with a length equal to or less than the max packet size.
  4) Checking the command length in the USB SCSI emulation.
  5) Checking command parameters within various emulations (SCSI (BBB), etc.)
  6) Implemented "Boot Protocol" for the (HID) Mouse emulation.
  7) Fixed wheel mouse HID report (along with the other HID reports)
  8) Added the Over Current bits to the UHCI code
  9) Added the monitoring of the toggle bit in the TDs

 Major:
  1) Re-wrote the UHCI stack processing so that control and bulk reclamation can be used.
     (This made it so Windows 7 now works)
  2) Adding function to the USB SCSI emulation.
  3) Enhanced the xHCI emulation to further check for errors in a driver.


fysnet:
Looks like the compiler used for the checks doesn't like the NULL in uhci_core.cc.  I don't know how to change that in the pull request. Please either point me in the right direction to modify the file, or simply change the NULL to 0 (the digit zero) and try again. If I close with comment, will it drop this pull request and allow me to modify it? Let's try.

stlintel:
> Looks like the compiler used for the checks doesn't like the NULL in uhci_core.cc. I don't know how to change that in the pull request. Please either point me in the right direction to modify the file, or simply change the NULL to 0 (the digit zero) and try again. If I close with comment, will it drop this pull request and allow me to modify it? Let's try.

The right way is to create a branch and then push to the branch.
Every update to the branch will automatically update pull request as well.
I believe you also could try top remote push to [fysnet:master](https://github.com/fysnet/Bochs) and it hopefully will update the pull request

stlintel:
BTW, the complain is not about NULL

uhci_core.cc:663:31: warning: converting to non-pointer type 'Bit32u' {aka 'unsigned int'} from NULL [-Wconversion-null]
  663 |     Bit32u item, queue_addr = NULL;

It says you assign NULL to non-pointer variable.

stlintel:
I see you already have 2 commits into pull request so you could continue adding more with the fixes

PR #30, opened 2023-02-17
USB Emulation enhancement and fixes

fysnet:
Minor:
  1) Check speed indicator in the Transfer Descriptor matching the device speed.
  2) Max packet size check, checking that the amount requested/sent is within limits.
  3) Checking that the first packet sent is the GetDescriptor Request with a length equal to or less than the max packet size.
  4) Checking the command length in the USB SCSI emulation.
  5) Checking command parameters within various emulations (SCSI (BBB), etc.)
  6) Implemented "Boot Protocol" for the (HID) Mouse emulation.
  7) Fixed wheel mouse HID report (along with the other HID reports)
  8) Added the Over Current bits to the UHCI code
  9) Added the monitoring of the toggle bit in the TDs

 Major:
  1) Re-wrote the UHCI stack processing so that control and bulk reclamation can be used.
     (This made it so Windows 7 now works)
  2) Adding function to the USB SCSI emulation.
  3) Enhanced the xHCI emulation to further check for errors in a driver.


fysnet:
Error'd again. I will try again.

stlintel:
```
usb_msd.cc: In member function 'virtual int usb_msd_device_c::handle_data(USBPacket*)':
usb_msd.cc:750:21: error: 'FALSE' was not declared in this scope
  750 |   bool was_status = FALSE; // so don't dump the status packet twice
      |                     ^~~~~
usb_msd.cc:852:24: error: 'TRUE' was not declared in this scope
  852 |           was_status = TRUE;
      |                        ^~~~
```

We are in C++, use true and false keywords

PR #42, opened 2023-04-01
Updates to USBMSD/SCSI emulation

fysnet:
- fixed default speed setting
  - Now checks for a valid speed of a device depending on the Host Controller as well as the device's allowed speeds.
  - fixed SCSI:Request Sense
  - fixed SCSI:Mode Sense
  - fixed SCSI:Inquiry
  - added SCSI:Get Configuration
  - added SCSI:Request Sense: ASC and ASCQ
  - USBMSD now handles short packets instead of returning s.datalen (s.scsi_len is now returned)
  - added the start of UASP Task Management
  - UASP: Return Status includes Sense
  - fixed xHCI:SET_TR_DEQUEUE command
  - fixed resetting of the TD's toggle bit
  - I have had all devices work on Win95, Win98, WinXP, Win7, Win8, and Win10, except for a few quirks with 
           Super-speed CD-ROMs, EHCI's UASP support, and num_lock states with HIDs.
  - Added more documentation to user.dbk
  To be fixed:
  - USBMSD write in WinXP still gives error. I think it is a timing issue. Write works in later Windows versions.

### DIFF
--- a/bochs/iodev/usb/scsi_device.cc
+++ b/bochs/iodev/usb/scsi_device.cc
@@ -8,6 +8,7 @@
 //  Based on code by Fabrice Bellard
 //
 //  Written by Paul Brook
+//  Updated and added to by Benjamin D Lunt (fys [at] fysnet [dot] net)
 //
 //  Copyright (C) 2007-2023  The Bochs Project
 //
@@ -84,11 +85,14 @@ scsi_device_t::scsi_device_t(device_image_t *_hdimage, int _tcq,
   hdimage = _hdimage;
   requests = NULL;
   sense = 0;
+  asc = 0;
+  ascq = 0;
   tcq = _tcq;
   completion = _completion;
   dev = _dev;
   block_size = hdimage->sect_size;
   locked = 0;
+  read_only = 0;
   inserted = 1;
   max_lba = (hdimage->hd_size / block_size) - 1;
   curr_lba = max_lba;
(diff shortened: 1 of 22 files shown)


bochs/iodev/usb/scsi_device.cc
```diff
@@ -1181,6 +1384,343 @@ Bit32s scsi_device_t::scsi_send_command(Bit32u tag, Bit8u *buf, Bit8u cmd_len, i
...
+      
+    // MMC3r10g, section 6.3.11, page 311(347)
+    case PAGE_CAPABILITIES:  // capabilities page
+      if (type == SCSIDEV_TYPE_CDROM) {
+        size = scsi_do_modepage_hdr(p, sub_page, page_code, 24);
+        p += size;
+        if (pc == PAGE_CONTROL_CHANGEABLE) {
+          memset(p, 0, 24);
+        } else {
+          p[ 0] = 0x03;                 // CD-R or CD-RW read/only
```
stlintel, line 1691: Ben, for future - I think it could be good idea to have defined struct (properly packed and etc) with all relevant fields, fill the struct on the side filed-by-field and then use memcpy to put into the buffer p

stlintel (approved):

bochs/iodev/usb/scsi_device.cc
fysnet: I agree, though the reason I did it this way, for now, is that some of the fields may change depending on the device it is associated with. If, later, I find that the fields will remain static, I can do as you suggest.  Even if only one of the fields change, as with this example, it will still be beneficial. Thank you.

PR #46, opened 2023-04-13
Fix zero-byte short-packet emulation

fysnet:
- This PR allows the host controller to process Control EP zero-byte packets in case the Guest doesn't specify to use short packets or an even number of bytes is transferred where the data length lands on a max packet size boundary.  The HC must still be allowed to process the next packet, even though it is a zero-length packet, to indicate a short-packet return. Before, the device went straight to the Status stage at the end of the data transfer. (Change effects Control transfers only)
- This PR changes the default speed to Full instead of Low if the speed is not specified.
- This PR allows a value of zero to be specified for a Configure request. A value of zero indicates to de-configure the device.

### DIFF
--- a/bochs/iodev/usb/usb_common.cc
+++ b/bochs/iodev/usb/usb_common.cc
@@ -288,6 +288,11 @@ usb_device_c *usbdev_locator_c::create(const char *type, const char *devname)
 #undef LOG_THIS
 #define LOG_THIS
 
+// Generic USB packet handler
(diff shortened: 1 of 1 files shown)


fysnet:
Maybe I still don't understand branches. This says it is a branch, but still states master->master. Sorry. I will investigate further.

fysnet:
I will close this PR and resubmit as a branch (hopefully).

PR #47, opened 2023-04-15
Fix zero-byte short-packet emulation

fysnet:
-This PR allows the host controller to process Control EP zero-byte packets in case the Guest doesn't specify to use short packets or an even number of bytes is transferred where the data length lands on a max packet size boundary. The HC must still be allowed to process the next packet, even though it is a zero-length packet, to indicate a short-packet return. Before, the device went straight to the Status stage at the end of the data transfer. (Change effects Control transfers only)
-This PR tries to change the default speed to Full, defaulting to d.minspeed, if the speed is not specified.
-This PR allows a value of zero to be specified for a Configure request. A value of zero indicates to de-configure the device.
-This PR fixes the Packet Capture format of a HID device

### DIFF
--- a/bochs/iodev/usb/usb_common.cc
+++ b/bochs/iodev/usb/usb_common.cc
@@ -49,22 +49,14 @@ bx_usbdev_ctl_c::bx_usbdev_ctl_c()
 
 void bx_usbdev_ctl_c::init(void)
 {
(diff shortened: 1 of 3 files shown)


fysnet:
Since this PR may cause contention with a current function of the source, I am retracting this PR.

PR #51, opened 2023-05-22
Fixes in VM Entry Checks for Guest Segment Registers

rei141:
This pull request includes two critical fixes in the VM entry checks related to the segment registers, ensuring better conformity with the Intel Software Developer's Manual.

1. Correction in checks for Code Segment (CS):
Previously, the DPL of CS was incorrectly compared with its own RPL. 
According to  Intel SDM, Vol. 3C, Chapter 27.3.1.2, for non-conforming code segments (type 9 or 11), the DPL of CS should be equal to the DPL of SS. 
And for conforming code segments (type 13 or 15), the DPL of CS cannot be greater than the DPL of SS. 

2. Correction in Type range checks for DS, ES, FS, GS:
The original code erroneously applied the check for types less than 11, excluding types equal to 11. 
This is not in accordance with Intel SDM, Vol. 3C, Chapter 27.3.1.2, which states that the check should include types equal to or less than 11. 
This fix corrects this by including types equal to or less than 11 in the check.

### DIFF
--- a/bochs/cpu/vmx.cc
+++ b/bochs/cpu/vmx.cc
@@ -1569,21 +1569,32 @@ Bit32u BX_CPU_C::VMenterLoadCheckGuestState(Bit64u *qualification)
      }
 
      if (n == BX_SEG_REG_CS) {
+        // Load SS
+        Bit16u ss_selector  = VMread16(VMCS_16BIT_GUEST_SS_SELECTOR);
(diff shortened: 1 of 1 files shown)


stlintel:
part 2 merged to main, verifying part 1

stlintel:
Merged into main with slightly different implementation.
Thanks !

PR #64, opened 2023-08-26
add port e9 hack support for all rings

therealdreg:
By enabling the 'all_rings' option, you can utilize the port e9 hack from ring3

IMO very useful for:
- user-mode sandbox (ex Cuckoo)
- malware analysis
- API/SYSCALL logger with a simple hook from ring3
- automation + instrumentation from user mode code
- ...

So yes, from this PR a user-mode-sandbox can display on the console of the system running Bochs anything that is written to 0xE9 port

porte9hackallrings

**This PR is 100% backward compatibility**

btw, @stlintel I'm not certain if **bochs/config.cc** is the ideal location to define **bool port_e9_hack_all_rings** (unmapped io/dev is better?)

### DIFF
--- a/bochs/config.cc
+++ b/bochs/config.cc
@@ -1679,6 +1679,13 @@ void bx_init_options()
       "Debug messages written to i/o port 0xE9 will be displayed on console",
       0);
 
+  // port e9 hack all rings
+  new bx_param_bool_c(misc,
+      "port_e9_hack_all_rings",
+      "Enable port 0xE9 hack for all rings",
+      "Debug messages written to i/o port 0xE9 from ring3 will be displayed on console",
+      0);
+
   // GDB stub
   menu = new bx_list_c(misc, "gdbstub", "GDB Stub Options");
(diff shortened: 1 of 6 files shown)


bochs/config.cc
```diff
@@ -62,6 +62,8 @@ int bochsrc_include_level = 0;
 
 extern bx_debug_t bx_dbg;
 
+bool port_e9_hack_all_rings;
```
stlintel, line 65: the config.cc is not ideal location and any other location is not ideal.
we are doing Bochs param tree exactly this stuff.
see for example for how MWAIT_IS_NOP handled.

You create new bx_param_bool_c with the value e9_all_ports and set it in config.cc
When read in io.cc

stlintel (changes requested):

bochs/config.cc
therealdreg: done!

therealdreg:
added PARAM_TREE.TXT, @stlintel take a look

stlintel (approved):

stlintel:
Merged to main

PR #66, opened 2023-08-27
add iodebug support for all rings complement #64

therealdreg:
This PR is for "I/O Interface to Bochs Debugger" from ring3 (port range: **0x8A00 - 0x8A01**)

PR #64 was for E9 HACK (port **0xE9**)

By enabling the iodebug's 'all_rings' option, you can utilize the port I/O Interface to Bochs Debugger from ring3. This PR allows the code running inside Bochs (ring3) to monitor memory ranges, trace individual instructions, and observe register values during execution.

https://bochs.sourceforge.io/doc/docbook/development/debugger-advanced.html

IMO very useful for:

- user-mode sandbox (ex Cuckoo)
- malware analysis
- API/SYSCALL hook/monitor from ring3
- automation + instrumentation from user mode code
...

A complement for PR #64

**This PR is 100% backward compatibility**

IODEBUG ALL RINGS

btw, @stlintel I'm not certain about:
- if misc is the ideal location for this.. should I create a new iodebug-option for this? 
- should I remove the #if in 'new bx_param_bool_c'?
- the new .bochsrc entry makes sense for you?

### DIFF
--- a/bochs/config.cc
+++ b/bochs/config.cc
@@ -1686,6 +1686,15 @@ void bx_init_options()
       "Debug messages written to i/o port 0xE9 from ring3 will be displayed on console",
       0);
 
+#if BX_SUPPORT_IODEBUG
+// iodebug all rings
+  new bx_param_bool_c(misc,
+      "iodebug_all_rings",
+      "Enable iodebug ports for all rings",
+      "I/O Interface to Bochs Debugger plugin for all rings",
+      0);
+#endif
+
   // GDB stub
   menu = new bx_list_c(misc, "gdbstub", "GDB Stub Options");
   menu->set_options(menu->SHOW_PARENT | menu->USE_BOX_TITLE);
@@ -1771,6 +1780,9 @@ void bx_init_options()
   misc->add(SIM->get_param(BXPN_USER_SHORTCUT));
   misc->add(SIM->get_param(BXPN_PORT_E9_HACK));
   misc->add(SIM->get_param(BXPN_PORT_E9_HACK_ALL_RINGS));
+  #if BX_SUPPORT_IODEBUG
+  misc->add(SIM->get_param(BXPN_IODEBUG_ALL_RINGS));
+  #endif
   misc->set_options(misc->SHOW_PARENT | misc->SHOW_GROUP_NAME);
 }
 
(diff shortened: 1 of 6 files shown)


bochs/config.cc
```diff
@@ -3221,6 +3233,21 @@ static int parse_line_formatted(const char *context, int num_params, char *param
     if (parse_port_e9_hack(context, (const char **)(params + 1), num_params - 1) < 0) {
       return -1;
     }
+  } else if (!strcmp(params[0], "iodebug")) {
+#if BX_SUPPORT_IODEBUG
+    if (num_params != 2) {
+      PARSE_ERR(("%s: iodebug directive: wrong # args.", context));
```
stlintel, line 3239: shouldn't it be the same as port_e9_hack ?
i.e. with enable/disable and all-rings ?
and here it could be parsed with the same method parse_port_e9_hack which will get extra param - path to param tree

bochs/cpu/io.cc
```diff
@@ -869,6 +869,12 @@ bool BX_CPP_AttrRegparmN(3) BX_CPU_C::allow_io(bxInstruction_c *i, Bit16u port,
   if (0xe9 == port && port_e9_hack_all_rings)
     return(1); // port e9 hack can be used by unprivileged code
 
+#if BX_SUPPORT_IODEBUG
+  static bool iodebug_all_rings = SIM->get_param_bool(BXPN_IODEBUG_ALL_RINGS)->get();
+  if ((0x8A00 == port || 0x8A01 == port) && iodebug_all_rings)
```
stlintel, line 874: 0x8A00 == (port & 0xfffe) ?

stlintel (changes requested):

bochs/config.cc
therealdreg: This Bochs plugin does not operate in that manner. It did not have a prior section in the .bochsrc file. Would you like me to create one? (with enabled= + all rings?)

bochs/cpu/io.cc
therealdreg: oka, done!

bochs/config.cc
therealdreg: I mean..  @stlintel, this plugin should be "disabled" with:

plugin_ctrl: iodebug=0

stlintel: Agree. Just wanted to avoid code duplication around config.cc

stlintel (approved):

stlintel:
This should just auto-push when you press 'resolve conversation' everywhere

therealdreg:
the automerge works, thx @stlintel !

PR #68, opened 2023-09-02
WIP: Say hello to test registers implementation

bog-dan-ro:
Implement https://sourceforge.net/p/bochs/feature-requests/502

### DIFF
--- a/bochs/cpu/decoder/disasm.cc
+++ b/bochs/cpu/decoder/disasm.cc
@@ -472,6 +472,12 @@ char *disasm_regref(char *disbufptr, const bxInstruction_c *i, unsigned src_num,
     disbufptr = dis_sprintf(disbufptr, "dr%d", srcreg);
     break;
 
+#if !BX_SUPPORT_X86_64
+  case BX_TREG:
+    disbufptr = dis_sprintf(disbufptr, "tr%d", srcreg);
+    break;
+#endif
+
   default:
     if (src_type != BX_NO_REGISTER)
       disbufptr = dis_sprintf(disbufptr, "(unknown source type %d)", src_type);
--- a/bochs/cpu/cpu.h
+++ b/bochs/cpu/cpu.h
@@ -904,6 +904,11 @@ public: // for now...
   bx_cr0_t   cr0;
   bx_address cr2;
   bx_address cr3;
+#if !BX_SUPPORT_X86_64
+  bx_tr6_t tr6;
+  bx_tr7_t tr7;
+#endif
+
 #if BX_CPU_LEVEL >= 5
   bx_cr4_t   cr4;
   Bit32u cr4_suppmask;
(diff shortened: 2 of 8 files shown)


bochs/cpu/decoder/disasm.cc
```diff
@@ -472,6 +472,10 @@ char *disasm_regref(char *disbufptr, const bxInstruction_c *i, unsigned src_num,
     disbufptr = dis_sprintf(disbufptr, "dr%d", srcreg);
     break;
 
+  case BX_TREG:
+    disbufptr = dis_sprintf(disbufptr, "tr%d", srcreg);
```
bog-dan-ro, line 476: For some reason this part doesn't work, so the (gui) debugger doesn't show the instructions correctly.
Can someone help me?

stlintel:
Wanna mention according to which spec you did it ?
I won't believe that 386 had 2^32 elements space for saving various tr7 values
this won't be merged ever due this reason

bochs/cpu/decoder/disasm.cc
stlintel: encodings above 0xF are not possible, there is assumption of 4 bit
in order to introduce a new encoding need to open a new space

stlintel:
Actually Bochs has TLB and if you want you could even implemented exact TR6/TR7 as for Cyrix docs you attached
Should not be too complicated

bochs/cpu/decoder/disasm.cc
bog-dan-ro: How do I do that?

bog-dan-ro:
> Actually Bochs has TLB and if you want you could even implemented exact TR6/TR7 as for Cyrix docs you attached Should not be too complicated

Partially done, I need  help with some attributes that I don't know how to set/get

bog-dan-ro:
I have some bad new on this matter.
The tr6,tr7 tests are 386 specific tests and they are giving different results on newer procs.
To test it I created a small C++ (watcom) test for DOS:
```
#include <iostream>
#include <stdint.h>

extern void set_tr7 (uint32_t tr7, uint32_t tr6);
#pragma aux set_tr7 = \
    ".386" \
    "mov tr7, edx" \
    "mov tr6, eax" \
    parm [edx] [eax];

extern uint32_t get_tr7 (uint32_t tr6);
#pragma aux get_tr7 = \
    ".386" \
    "mov tr6, edx" \
    "mov eax, tr7" \
    parm  [edx] \
    value [eax];

using namespace std;
int main()
{
    uint32_t tr7 = 0x10000010;
    uint32_t tr6 = 0xaa0;
    for (int bl = 0; bl < 4; ++bl) {
        for (int ecx = 0; ecx < 8; ++ecx) {
            set_tr7(tr7, tr6);
            tr7 += 0x1000;
            tr6 += 0x1000;
        }
        tr7 += 4;
    }
    tr7 = 0x10000010;
    tr6 = 0xaa1;
    for (int bl = 0; bl < 4; ++bl) {
        for (int ecx = 0; ecx < 8; ++ecx) {
            const uint32_t rtr7 = get_tr7(tr6);
            cout << "tr7 = 0x" << hex << rtr7 << " == 0x" << tr7 << endl;
            tr7 += 0x1000;
            tr6 += 0x1000;
        }
        tr7 += 4;
    }
    return 0;
}
```

On 386 it pass the tests, but on 486 it doesn't, as predicted here https://www.scs.stanford.edu/05au-cs240c/lab/i386/s10_06.htm .

I'm going to close this PR.

stlintel:
I bet having these MOV TR6/7 in 38 as NOP (instead of UD) will be good enough and pass the tests

bog-dan-ro:
> I bet having these MOV TR6/7 in 386 as NOP (instead of UD) will be good enough and pass the tests But anyway - now you see why they are not implemented in Bochs :)

It won't work. The BIOS code that I was talking about looks like this:
```

PR #72, opened 2023-09-06
Fix zero length packet handling

fysnet:
This fixes zero length packet handling.
See the following example that requests 128 bytes:
```
SETUP(8)
IN(64)
IN(64)
STATUS(0)
```
The current code erroneously assumes that there are no more IN packets after the first two 64-byte packets.

However, what happens with the following:
```
SETUP(8)
IN(64)
IN(64)
IN(64)    <--- current code assumes this will be, and expects a STATUS packet
STATUS(0)
```
Currently, the third IN(64) above will result in a coding error because the code is expecting the STATUS packet, not another IN packet.

The "controller" must allow for more packets than expected, returning a short packet detect on the third IN(64) packet shown above, actually returning zero bytes, hence the Short Packet Detect.

This patch was tested on WinXP, Win7, and Win10.

### DIFF
--- a/bochs/iodev/usb/usb_common.cc
+++ b/bochs/iodev/usb/usb_common.cc
@@ -49,22 +49,14 @@ bx_usbdev_ctl_c::bx_usbdev_ctl_c()
 
 void bx_usbdev_ctl_c::init(void)
 {
-  Bit8u i, j, count;
-
-  /*  If you wish to set DEBUG=report in the code, instead of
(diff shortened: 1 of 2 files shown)


stlintel:
Why did you closed and deleted this ? It was in purpose ?

fysnet:
It was the wrong two files. I have a set of actual code, and a set of experimental code and I accidentally posted the experimental code here. PR #73 is the correct set of files.
Thank you,
Ben

PR #82, opened 2023-10-05
Fix OHCI Toggle function

fysnet:
This fixes the toggle bit check in the OHCI.
Tested Guests: WinXP, Win7
Tested devices: HID, HUB, DISK, CD-ROM


stlintel (approved):

fysnet:
This pull request seems to be missing from the Bochs tree even after you committed it 2 days ago? I restored the branch, but I don't think that is the issue.

stlintel:
Very weird, pushed it again

PR #83, opened 2023-10-07
Ohci toggle

fysnet:
For some unknown reason, PR #82 was committed but it did not stick, so here is another try.


fysnet:
Looks like PR #82 was re-merged 9 hours ago. Unknown if this PR is needed.

PR #84, opened 2023-10-07
implemented SM3 instructions

stlintel:
add rol/ror methods to scalar_arith.h and use in more places


Nable80:
Is it OK to mix style and functional changes in one PR? Style changes look completely unrelated and I'm afraid they are even breaking ones, I definitely remember ``make`` treating tabs and spaces differently.

bochs/cpu/sm4.cc
```diff
@@ -0,0 +1,162 @@
...
+#include "cpu.h"
+#define LOG_THIS BX_CPU_THIS_PTR
+
+#if BX_SUPPORT_AVX
+
+#include "scalar_arith.h"
+
+// Additional details can be found at: https://tools.ietf.org/html/draft-ribose-cfrg-sm4-10
+
+static Bit8u sm4_sbox[256] = {
```
Nable80, line 35: ``static const``?

stlintel: yep, good idea

fysnet:
This PR breaks Win10 booting.

Line 231 in scalar_arith.h
```
BX_CPP_INLINE Bit32u ror64(Bit32u v64, unsigned count)
{
  return (v64 >> count) | (v64 << (64 - count));
}
```
returns a 32-bit value when it should return a 64-bit value.

Most likely a copy/paste oversight.

fysnet:
I even missed the second Bit32u. :-)
Anyway, thank you for all you guys do.

PR #104, opened 2023-10-20
Ehci companion

fysnet:
This PR adds the ability to choose which companion controller the EHCI uses.
`#usb_ehci: enabled=1, companion=uhci`
`#usb_ehci: enabled=1, companion=ohci`

UHCI is enabled by default:
`#usb_ehci: enabled=1`

This breaks up `usb_ohci.cc/h` into `usb_ohci.cc/h` and `ohci_core.cc/h` to be like the existing `usb_uhci.cc/h` and `uhci_core.cc/h`.
This adds the ability to define a Port Routing string in the `HcspPortRoute` register, and is optional via the `EHCI_PORT_ROUTE` define.

Currently `EHCI_N_CC` and `EHCI_N_PCC` must remain as currently set (3 and 2 respectively), however with a little more work, I think these can be modified and have up to 7 companion controllers with more ports per companion, as long as the combined total is <= 15 total ports.

Please verify the `Makefile.in` changes. I do not use GCC and can not verify it is correct.


stlintel:
You actually do verify it with CI/DI here.
Every pull request runs CI/DI checks with 3 configure scripts - .conf.nothing, .conf.everything, release configuration
The 'all checks passed' means you did all changes fine and everything compiles just fine.

PR #109, opened 2023-10-23
Add Modeless MessageBox capabilities

fysnet:
As an example, when saving the state of an emulation, especially one that has large image files and a large memory setting, there is a pause between the time you choose a directory to save to, and the MessageBox that states that the state was saved. During this time, there is no indication that Bochs is doing anything. As far as the user is concerned, Bochs froze.

So, this PR adds the capability to display a _Modeless_ message box. To do so, at any point you need a message box shown, allowing other things to continue to happen (SYNCHRONOUS), simply call the following line:
`void *hwnd = SIM->ml_message_box("Message Box Title", "Message Box Text");`
Then after the task has been performed, remove the message box:
`SIM->ml_message_box_kill(hwnd);`

The function is implemented in all configurations via virtual calls, however, the creation of the message box is currently only implemented in the Win32 configuration.  Other configurations will need this implemented, though I do not have those configurations to test with, except the text only config. I may get around to it :-).

This also currently only displays as the state is being saved. To display when the state is being restored should now be a simple task by adding the two lines shown above to the associated function.

The only drawback with this is that if the state is considerably small, the save will take just a second or less. If so, this window will show and then instantly be removed, possibly confusing the user. _What was that!!_

### DIFF
--- a/bochs/gui/gui.cc
+++ b/bochs/gui/gui.cc
@@ -819,17 +819,19 @@ void bx_gui_c::userbutton_handler(void)
 
 void bx_gui_c::save_restore_handler(void)
 {
-  int ret;
   char sr_path[BX_PATHNAME_LEN];
 
   if (BX_GUI_THIS dialog_caps & BX_GUI_DLG_SAVE_RESTORE) {
     BX_GUI_THIS set_display_mode(DISP_MODE_CONFIG);
     sr_path[0] = 0;
-    ret = SIM->ask_filename(sr_path, sizeof(sr_path),
+    int ret = SIM->ask_filename(sr_path, sizeof(sr_path),
                             "Save Bochs state to folder...", "none",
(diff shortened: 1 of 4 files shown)


stlintel:
Need to test it carefully for each GUI library (X, win32, etc)

vruppert:
I can confirm that the new feature works in Windows, but I would prefer using a modeless dialog box instead. As a stub for the X gui, these lines should be added in textconfig.cc:

    case BX_SYNC_EVT_ML_MSG_BOX:
      fprintf(stderr, "%s\n%s\n", event->u.logmsg.prefix, event->u.logmsg.msg);
      return event;
    case BX_SYNC_EVT_ML_MSG_BOX_KILL:
      // Nothing to do
      return event;

I'll have a look how to implement this feature in the x and wx guis.

vruppert (approved):

PR #110, opened 2023-10-23
write companion to save file

fysnet:
This fix adds the `companion=uhci\ohci` parameter to the saved config file on suspend.

### DIFF
--- a/bochs/config.cc
+++ b/bochs/config.cc
@@ -3394,12 +3394,14 @@ int bx_write_floppy_options(FILE *fp, int drive)
 #if BX_SUPPORT_PCIUSB
 int bx_write_usb_options(FILE *fp, int maxports, bx_list_c *base)
 {
-  int i;
   char tmpname[24], tmpstr[BX_PATHNAME_LEN];
 
   fprintf(fp, "usb_%s: enabled=%d", base->get_name(), SIM->get_param_bool("enabled", base)->get());
+  // if we are the ehci, we need to add the companion= parameter
+  if (!strcmp(base->get_name(), "ehci"))
+    fprintf(fp, ", companion=%s", (SIM->get_param_enum(BXPN_EHCI_COMPANION)->get() == 1) ? "ohci" : "uhci");
   if (SIM->get_param_bool("enabled", base)->get()) {
-    for (i = 1; i <= maxports; i++) {
+    for (int i = 1; i <= maxports; i++) {
       sprintf(tmpname, "port%d.device", i);
       SIM->get_param_enum(tmpname, base)->dump_param(tmpstr, BX_PATHNAME_LEN, 1);
       fprintf(fp, ", port%d=%s", i, tmpstr);


stlintel (approved):

stlintel:
Ideally I would like all options to be saved using param tree with simple bx_write_param_list call
and read using bx_parse_param_from_list call
this way to parameter will be ever forgotten

fysnet:
Please see PR #111 to see if this is better.

stlintel:
No, you manually write parameters instead of letting param tree to do the job for you

fysnet:
Then I don't understand what you mean. Sorry.

stlintel:
Look how KEYBOARD options written to config from the same function. Or MOUSE. Or new created E9_PORT_HACK. Or CPU.
It is just single function call like:
  bx_write_param_list(fp, (bx_list_c*) SIM->get_param(BXPN_KEYBOARD), NULL, 0);
The options are defined as hierarchical structure in param_tree.
For E9_PORT_HACK for example there is base param bx_list_c named "e9_port_hack" and it has 2 children 'enabled' and ''all_rings'. The params "know" their type and know to read themselves and write themselves.
So param tree could write the .bochsrc by single function call.
Similarly it could read all parameters using single function call.
See how MOUSE params are read from .bochsrc:

```
  } else if (!strcmp(params[0], "mouse")) {
    if (num_params < 2) {
      PARSE_ERR(("%s: mouse directive malformed.", context));
    }
    for (i=1; i<num_params; i++) {
      if (bx_parse_param_from_list(context, params[i], (bx_list_c*) SIM->get_param(BXPN_MOUSE)) < 0) {
        PARSE_ERR(("%s: mouse directive malformed.", context));
      }
    }

```

This is generic. If you add more params you shouldn't touch this code. Param tree know how to read it and also how to write it.
Ideally I would like to replace entire bx_write_configuration to a single function call, smth like:

`  bx_write_param_list(fp, (bx_list_c*) SIM->get_param(BX_PARAM_ROOT), NULL, 0);`

and it will recursively write all options will all params.

fysnet:
I see what you mean now. I wasn't sure since when I did that before, using:

`bx_write_param_list(fp, (bx_list_c*) SIM->get_param(BXPN_USB_EHCI), NULL, 0);`

it produced:

`ehci: enabled=true, companion=ohci, model=uPD720202, n_ports=-1, port1=, port2=, port3=, port4=, port5=, port6=`

which has many errors to it. `ehci` should be `usb_ehci`, which I guess could be fixed with:

`bx_write_param_list(fp, (bx_list_c*) SIM->get_param(BXPN_USB_EHCI), "usb_ehci", 0);`

But that defeats the purpose intended since now you have to provide a hard coded string.
It also includes some of the xHCI stuff *and* -1 is a bad `n_ports` value anyway.

I am guessing you mean that there needs to be some work done not only replacing `bx_write_usb_options`, but other work so that eventually we can simply call `bx_write_param_list`.

Add it to the todo list, right? :-)

Issue #123, opened 2023-11-07
Panic while formatting floppy in Windows 2000


Vort:
When I execute command `format a:` in Windows 2000 _(or Windows XP)_, Bochs shows error `could not perform lseek() to -512 on floppy image file`.

![bochs_format_panic](https://github.com/bochs-emu/Bochs/assets/1242858/ffdbddd7-52db-4efa-a970-c6e82fbadcd8)

Version: 75d076b990a5a8198b6bf9a8ee1e43a37483eba3.

fysnet:
I had some issues with the FDC code on older Guests, so I went investigating. I found a few errors and one thing led to another and I have a considerable update for the FDC emulation. The Format function was my next focus. I will look into this.

Thanks.

fysnet:
I can confirm this error. However, with my new updates, this error has been remedied. As soon as I do some more testing and verify that my new updates work on all disk types and sizes, I will post a PR.

Vort:
@fysnet do you know about OmniFlop program? May be useful for testing.

fysnet:
I had not. I will have a look. Thank you.

fysnet:
@Vort, please have a look at PR #135. It drastically enhances the floppy emulation, which should fix the issue you show here.
While you are at it, please test the new emulation and let me know if you find any errors. 
I tested it with many guests in many instances and different types and disk sizes, however, since it is a drastic change, all the testing I can get will help.

Vort:
> should fix the issue you show here

I confirm fix for `lseek() to -512` problem, thanks.

> please test the new emulation and let me know if you find any errors

I'm not that experienced to test properly such huge change.
But I made few tests and found two problems.
1. Not related to your changes, it was like that before. If I press Reset button in UI while floppy is accessed, "green light" stays on after restart.
2. Partially related to your changes. I tried to use 360K floppy in 1.2M drive. DOS was working fine with it, but when I booted to WinXP, and tried to access floppy, OS displayed controller error or something like that. Before your changes, it asked to format floppy - not better. I decided to check this scenario on real hardware: floppy drive was producing weird noises for quite time, but eventually WinXP showed correct files on disk.

fysnet:
1. I can confirm this. It should be a quick fix in the gui code. I will see about adding a PR for it.
[...]

PR #136, opened 2023-11-17
reset status bar items at reset

fysnet:
When the user presses the "Reset" button on the gui ribbon, we need to clear the active status of all statusbar items or they remain set after the reset occurs.
This is mentioned in issue #123

### DIFF
--- a/bochs/gui/gui.cc
+++ b/bochs/gui/gui.cc
@@ -420,6 +420,7 @@ void bx_gui_c::cdrom1_handler(void)
 void bx_gui_c::reset_handler(void)
 {
   BX_INFO(("system RESET callback"));
+  BX_GUI_THIS statusbar_setall(0);
   bx_pc_system.Reset(BX_RESET_HARDWARE);
 }
 
@@ -984,6 +985,12 @@ void bx_gui_c::statusbar_setitem(int element, bool active, bool w)
       statusitem[element].counter = 5;
     }
   }
+}
+
+void bx_gui_c::statusbar_setall(bool active)
+{
+  for (int i=0; i<BX_MAX_STATUSITEMS; i++)
+    statusbar_setitem(i, active);
 }
 
 void bx_gui_c::led_timer_handler(void *this_ptr)
--- a/bochs/gui/gui.h
+++ b/bochs/gui/gui.h
@@ -192,6 +192,7 @@ public:
   int register_statusitem(const char *text, bool auto_off=0);
   void unregister_statusitem(int id);
   void statusbar_setitem(int element, bool active, bool w=0);
+  void statusbar_setall(bool active);
   static void init_signal_handlers();
   static void toggle_mouse_enable(void);
   bool mouse_toggle_check(Bit32u key, bool pressed);


Vort:
I confirm that floppy "LED" now turns off with Reset.

However, it looks like behaviour of Num Lock became worse:
If I enable Num Lock and press Reset, LED turns off.
But when I press Num Lock afterwards, LED does not turn on.
However, actual function of Num Lock becomes activated.

stlintel (approved):

stlintel:
Sorry, didn't notice comment by Vort before :(

Vort:
> I suggest may be to touch only CDROM-HDD-FDD LEDs

Num Lock was not working correctly before - only slightly better.
So it needs to be fixed anyway.

fysnet:
Shouldn't the reset code set the Num Lock, Caps Lock, and Scroll lock to their respective reset states? What if the user presses the caps lock, then presses the reset button. Does the caps lock stay set? Yes, so this is in error.
The reset code needs to be modified to set these three to their respective reset states.
Therefore, in my opinion, this PR needs to be accepted as it was originally posted and the reset code needs to be modified.

Vort:
It is possible to discuss which bug is better, but I think it is better to just get rid of them.
@fysnet if [my explanation](https://github.com/bochs-emu/Bochs/pull/136#issuecomment-1815802621) of what is wrong is not enough, I can explain better.

stlintel:
> However, it looks like behaviour of Num Lock became worse:
> If I enable Num Lock and press Reset, LED turns off.
> But when I press Num Lock afterwards, LED does not turn on.
> However, actual function of Num Lock becomes activated.

Yes, when one press Reset the LEDs should clear
But if so - the keyboard code should be changed too to deactivate the actual NumLock/CapsLock function on reset

```
If I enable Num Lock and press Reset, LED turns off.
But when I press Num Lock afterwards, LED does not turn on.
However, actual function of Num Lock becomes activated.
```

Vort:
This is what I meant, but was not able to say properly :/

fysnet:
Here is my thought on all of this.

The emulated guest will start with the NUM LOCK, SCRL LOCK, and CAPS LOCK all off. This is because the rombios.c sets these bits at 0x0417 to zero. (rombios.cc [line 10985](https://github.com/bochs-emu/Bochs/blob/master/bochs/bios/rombios.c#L10985)).

This is done with no regard to if the HOST has any of these in the active state. I.e.: most HOSTs will have the NUM LOCK in the active state.

Therefore, in the GUEST, the NUM LOCK will be inactive. To activate the NUM LOCK, you press the NUM LOCK key. The GUEST now sees the NUM LOCK active. However, the NUM LOCK is inactive on the HOST as well as the HOST's LED is off. This is the same for the other two keys. The correct scan codes are sent to the GUEST, but the HOST's LED doesn't match the GUEST's LED. Again, same for the other two keys.

If you have pressed the NUM LOCK key so that the Guest shows it active, then click on reset, the Guest, due to the bios code line 10985, will now have the NUM LOCK inactive.  If we don't clear the Guest's LED using this PR, this LED and the state of the numlock are not in sync. Again, this goes for all three key states.

To fix this, the rombios, at line 10985 should check the HOSTs key status and set that byte accordingly.  However, there is a problem with this. The rombios.cc code has no standard way of calling the HOST to get the current key status, especially if more than one HOST platform is supported.

Most HOSTs will have the NUM LOCK active, and the SCRL LOCK and CAPS LOCK inactive when they start a GUEST. Therefore, if at line 10985 in the BIOS, we set `al` to `0x20` instead of `0x00`, the GUEST should start with the NUM LOCK active. There would need to be a line at the end of `init()` after `DEV_register_default_keyboard(...` to set the LED to active. `DEV_kbd_set_indicator(0, BX_KBD_LED_NUM, true);`

In the keyboard.cc::reset() you would add `DEV_kbd_set_indicator(0, BX_KBD_LED_NUM, true);` also.

The only way to use the current code is to ignore the HOST's LEDs when using the GUEST. The problem with this is if you have more than one app open on the HOST and activate another app, the state of the CAPS LOCK and other two keys may be undesired.

aS AN EXAMPLE, i HAVE bOCHS RUNNING NOW, WITH THE caps lock ACTIVE IN THE guest AND MOVED TO HERE TO WRITE. aS YOU CAN SEE, CAPS LOCK IS NOW IN AN UNDESIRED STATE.

Moving back and forth between this text and the GUEST, pressing the CAPS LOCK each time toggles the HOST's state and LED, but does not toggle the GUEST's state and LED unless the BOCHS window is the active window.

My only solution is to be mindful of the state of these three keys while Bochs is running, using the status bar indicators as the LED status for the GUEST and the HOST's LEDs for everything else. Therefore, I believe this PR should be reinstated as is, and other than the rombios.cc modification mention above, no other change should be made.

Someone else may have a better solution, but this is my thought on the subject.

Ben

fysnet:
One other thought I had was that the HOST could send the key press of either of the three keys when they were pressed, and then ignore the key press itself, not toggling the host's status.

Vort:
> The only way to use the current code is to ignore the HOST's LEDs when using the GUEST.

Yeah, that's why I wrote "LED" in quotes, meaning virtual LEDs, inside of guest.
Tracking host's real LEDs is, most likely, complex task, which is beyond the scope of this simple fix.

> My only solution is to be mindful of the state of these three keys while Bochs is running, using the status bar indicators as the LED status for the GUEST

Problem is virtual LEDs in status bar works incorrectly.
Please reread what I initially posted considering that I was talking not about real LEDs, but about virtual ones.

fysnet:
Sorry, I have been away for a little while, just now returning and have to apologize. I read your previous post and my mind went somewhere else with it, at my fault. I now understand what you are saying. Sorry. I will have a look at it.

fysnet:
Okay, again, sorry for the misunderstanding.

Try the following additional line in the reset() function:
```
void bx_keyb_c::reset(unsigned type)
{
  BX_KEY_THIS s.kbd_internal_buffer.led_status = 0;
}
```
Adding that single line fixes it for me.

PR #138, opened 2023-11-17
initial code for UINTR implementation

stlintel:
First step into implementing UINTR - User Level Interrupts ISA extension
To be continued


bochs/configure.ac
```diff
@@ -1252,6 +1252,29 @@ if test "$support_cet" -gt 0 -a "$use_x86_64" = 0; then
...
+   fi
+   ],
+  [
+    AC_MSG_RESULT(no)
+    AC_DEFINE(BX_SUPPORT_UINTR, 0)
+  ]
+  )
+
+if test "$support_uintr" -gt 0 -a "$use_x86_64" = 0; then
+  AC_MSG_ERROR([User Level Interrupt feature emulation require x86-64 enabled])
```
Nable80, line 1275: require**s**?
I guess it was copied from another description...

bochs/cpu/cpu.h
```diff
@@ -954,6 +954,21 @@ class BOCHSAPI BX_CPU_C : public logfunctions {
   Bit32u wr_pkey[16];
 #endif
 
+#if BX_SUPPORT_UINTR && BX_SUPPORT_X86_64
```
Nable80, line 957: Is there any need to check ``BX_SUPPORT_X86_64`` here? I thought ``BX_SUPPORT_UINTR`` can't be enabled without it anyway.

bochs/cpu/uintr.cc
```diff
@@ -0,0 +1,291 @@
...
+
+  // ----------- UPID format --------
+  // |     00 | Outstanding notification (UPID.ON)
+  // |     01 | Suppress notification    (UPID.SN)
+  // |  15:02 | Reserved, MBZ
+  // |  23:16 | Notification vector, 8-bit
+  // |  31:24 | Reserved, MBZ
+  // |  63:32 | Notification destination, 32-bit APIC/X2APIC ID
+  // | 127:64 | Posted interrupt requests (PIR)
+  // |        | One bit for ech user-interrupt vector
```
Nable80, line 204: e**a**ch

bochs/cpu/vmx.cc
```diff
@@ -2122,10 +2133,19 @@ Bit32u BX_CPU_C::VMenterLoadCheckGuestState(Bit64u *qualification)
...
+  }
+#endif
+
 #if BX_SUPPORT_VMX >= 2
   if (vmentry_ctrls & VMX_VMENTRY_CTRL1_LOAD_PAT_MSR) {
     BX_CPU_THIS_PTR msr.pat = guest.pat_msr;
   }
+#endif
+
+#if BX_SUPPORT_VMX >= 2
```
Nable80, line 2148: Two adjacent blocks guarded by the same ``#if`` condition look suspicious.

bochs/cpu/xsave.cc
```diff
@@ -966,6 +966,83 @@ bool BX_CPU_C::xsave_cet_s_state_xinuse(void)
...
+  Bit64u ui_handler   = read_virtual_qword(i->seg(),  offset);
+  Bit64u stack_adjust = read_virtual_qword(i->seg(), (offset +  8) & asize_mask);
+  Bit64u misc         = read_virtual_qword(i->seg(), (offset + 16) & asize_mask);
+  bool UIF = (misc >> 63) != 0;
+  misc &= ~BX_CONST64(0x8000000000000000); // clear UIF
+  Bit64u upid_addr    = read_virtual_qword(i->seg(), (offset + 24) & asize_mask);
+  Bit64u uirr         = read_virtual_qword(i->seg(), (offset + 32) & asize_mask);
+  Bit64u uitt_addr    = read_virtual_qword(i->seg(), (offset + 40) & asize_mask);
+
+  // XRSTOR on UINTR state does all reserved bits and canonicality check like WRMSR would do
```
Nable80, line 1009: check**s**?

bochs/configure.ac
```diff
@@ -1252,6 +1252,29 @@ if test "$support_cet" -gt 0 -a "$use_x86_64" = 0; then
...
 fi
 
+support_uintr=0
+AC_MSG_CHECKING(for User Level Interrupt support)
+AC_ARG_ENABLE(cet,
+  AS_HELP_STRING([--enable-uintr], [for User Level Interrupt support (no)]),
+  [if test "$enableval" = yes; then
+    AC_MSG_RESULT(yes)
+    AC_DEFINE(BX_SUPPORT_UINTR, 1)
+    support_cet=1
```
Nable80, line 1262: Should it be ``support_uintr=1`` instead of ``support_cet=1``?

bochs/configure.ac
stlintel: Really required. The UINTR support ONLY on processor with X86_64 and ONLY in long64 mode.
There are two ways to enforce that.
I started from first one by adding everywhere 
`#if BX_SUPPORT_UINTR && BX_SUPPORT_X86_64`
but turned so ugly so I going away from that and prefer to enforce it in configure script + compile time assert in config.h

bochs/cpu/cpu.h
stlintel: I started with attempt to avoid configure time restrictions but recently changed my mind.
Will cleanup all these && BX_SUPPORT_X86_64 later

bochs/cpu/vmx.cc
stlintel: But they related to disjoint feature therefore I split them in purpose

bochs/cpu/xsave.cc
stlintel: checks are inside wrmsr calls

bochs/configure.ac
stlintel: Thanks !

bochs/configure
```diff
@@ -24299,6 +24137,40 @@ if test "$support_cet" -gt 0 -a "$use_x86_64" = 0; then
   as_fn_error $? "Control Flow Enforcement Technology feature emulation require x86-64 enabled" "$LINENO" 5
 fi
 
+support_uintr=0
+{ printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for User Level Interrupt support" >&5
+printf %s "checking for User Level Interrupt support... " >&6; }
```
Nable80, line 24142: Did you regenerate ``configure`` script in 63fe1d4429f0f8e191f09b03c912c802a646f1cf or just edited it manually? I see suspicious leftovers of ``enable-cet``/``enable_cet`` in UINTR section (lines 24142-24146).

bochs/configure.ac
Nable80: I meant a bit different matter: the phrase is about a single feature/single technology emulation, so the **-s** ending seems to be required for verbs in this context. I.e. "CET feature emulation require**s** x86-64 enabled", "UINTR feature emulation require**s** x86-64 enabled" and so on. Am I wrong here?

Sorry for grammar comments instead of technical ones.

bochs/configure
stlintel: Thanks, fixed

stlintel:
This branch seems ready. 
Would be glad of someone could review against SDM or simply look on the code and comment.
These comments are very useful and really help to get better code !

bochs/cpu/cpu.h
Nable80: @stlintel 
Will you include this cleanup into the same PR or it'll be really later?

bochs/cpu/event.cc
```diff
@@ -129,6 +129,8 @@ void BX_CPU_C::InterruptAcknowledge(void)
   }
 #endif
 
+  BX_CPU_THIS_PTR EXT = 1; /* external event */
```
Nable80, line 132: I'm sorry but I couldn't understand why this line was moved above.

bochs/cpu/event.cc
```diff
@@ -138,14 +140,25 @@ void BX_CPU_C::InterruptAcknowledge(void)
...
-  BX_CPU_THIS_PTR EXT = 1; /* external event */
 #if BX_SUPPORT_VMX
   VMexit_Event(BX_EXTERNAL_INTERRUPT, vector, 0, 0);
 #endif
 
-  BX_INSTR_HWINTERRUPT(BX_CPU_ID, vector,
+#if BX_SUPPORT_UINTR
+  if (BX_CPU_THIS_PTR cr4.get_UINTR() && long64_mode() && vector == BX_CPU_THIS_PTR uintr.uinv)
+  {
+#if BX_SUPPORT_APIC
```
Nable80, line 150: Is it actually possible to have a disabled ``BX_SUPPORT_APIC`` flag when ``BX_SUPPORT_UINTR`` is enabled?

Maybe ``config.h.in`` should contain additional sanity checks.

bochs/cpu/flag_ctrl_pro.cc
```diff
@@ -129,25 +129,32 @@ void BX_CPU_C::handleInterruptMaskChange(void)
...
+#endif
      return;
   }
 #endif
 
 #if BX_SUPPORT_SVM
   if (BX_CPU_THIS_PTR in_svm_guest && SVM_V_INTR_MASKING) {
      if (! SVM_HOST_IF)
-       mask_event(BX_EVENT_PENDING_INTR | BX_EVENT_PENDING_LAPIC_INTR);
+       mask_event(BX_EVENT_PENDING_INTR | BX_EVENT_PENDING_LAPIC_INTR | BX_EVENT_PENDING_UINTR);
```
Nable80, line 157: I thought UINTR is an Intel-only technology, how could it affect SVM-specific code?

bochs/cpu/uintr.cc
```diff
@@ -0,0 +1,344 @@
...
+    RSP  = BX_CPU_THIS_PTR uintr.stack_adjust;
+  else
+    RSP -= BX_CPU_THIS_PTR uintr.stack_adjust;
+  RSP &= BX_CONST64(0xfffffffffffffff0);
+
+  push_64(tmp_RSP);
+  push_64(read_eflags());
+  push_64(RIP);
+  unsigned vector = lzcntq(BX_CPU_THIS_PTR uintr.uirr); // find #most significant bit in UIRR
+  // expected to be not 0
```
Nable80, line 55: This comment looks a bit misleading in its context, i.e. what is expected to be non-zero? ``vector`` == 0 is OK after all. A check for ``uintr.uirr`` != 0 can be added to the initial BX_ASSERT if necessary.

bochs/cpu/uintr.cc
```diff
@@ -0,0 +1,344 @@
...
+  if (BX_CPU_THIS_PTR uintr.stack_adjust & 0x1)
+    RSP  = BX_CPU_THIS_PTR uintr.stack_adjust;
+  else
+    RSP -= BX_CPU_THIS_PTR uintr.stack_adjust;
+  RSP &= BX_CONST64(0xfffffffffffffff0);
+
+  push_64(tmp_RSP);
+  push_64(read_eflags());
+  push_64(RIP);
+  unsigned vector = lzcntq(BX_CPU_THIS_PTR uintr.uirr); // find #most significant bit in UIRR
```
Nable80, line 54: I expected smth like ``63 - lzcntq(…)`` here, am I wrong?

bochs/cpu/uintr.cc
Nable80: It became ``ech`` again, the latest commit didn't include this fix…

Nable80:
I couldn't find how to unmark resolved conversations, please check my comments there too.

Btw, how do you trigger ``#UD`` when UINTR instructions are attempted to use in non-64-bit modes, is ``ATTR_IS64`` enough? I couldn't understand this part of Bochs yet.

bochs/cpu/flag_ctrl_pro.cc
stlintel: I bet AMD will copy it soon so enabled it already as well as they usually do.
User-level interrupt is being asked by customers from both companies for very long time

bochs/cpu/uintr.cc
stlintel: No, you are absolutely right. Forgot to commit this part ...
Fixed now

bochs/cpu/uintr.cc
stlintel: Fixed comment and added assert

bochs/cpu/event.cc
stlintel: Today it is theoretically possible.
Moreover legacy APIC is going to be deprecated by Intel and only X2APIC will survive.
So even if today BX_SUPPORT_APIC=0 makes no sense it might become reasonable soon.

bochs/cpu/event.cc
stlintel: Actually no particular reason. This has no change at all.
Originally I have rewritten this block completely but undone the change.
But seems like undone not 100% exactly

stlintel:
>Btw, how do you trigger #UD when UINTR instructions are attempted to use in non-64-bit modes, is ATTR_IS64 enough? I couldn't understand this part of Bochs yet.

Yes, it is enough, The decoder matching instruction "mode" patterns and it "mode" doesn't match it won't match the pattern.
So if ATTR_IS64 was specified the instruction pattern won't match outside of 64-bit long mode.
And if no pattern will match - it will result in #UD.

bochs/cpu/cpu.h
stlintel: A bit later. Now I am trying to make some assembly tests to see that it works as expected.

PR #139, opened 2023-11-19
add keylocker feature implementation

stlintel:


bochs/configure.ac
```diff
@@ -1252,6 +1252,23 @@ if test "$support_cet" -gt 0 -a "$use_x86_64" = 0; then
   AC_MSG_ERROR([Control Flow Enforcement Technology feature emulation require x86-64 enabled])
 fi
 
+AC_MSG_CHECKING(for KeyLocker support)
+AC_ARG_ENABLE(3dnow,
```
Vort, line 1256: Is `3dnow` related to this change? If not, looks like mistake.

stlintel: Thanks, will fix !

PR #142, opened 2023-11-20
reset status bar items at reset

fysnet:
When the user presses the "Reset" button on the gui ribbon, we need to clear the active status of all statusbar items or they remain set after the reset occurs.
This is mentioned in issue https://github.com/bochs-emu/Bochs/issues/136


Vort:
With c91653a2001dcb57386cc56edfe22dca0d217099 both keyboard and floppy virtual LEDs behave correctly after Reset. Thank you.

PR #162, opened 2023-12-07
Added save/restore support for the ESCD space in the flash BIOS ROM.

vruppert:
These changes only have an effect when using a BIOS image designed for real hardware (tested with asus_p6np5.bin). When using Bochs BIOS no data will be saved / restored, since the 'flash_modified' value is always set to 'false'. Saving ESCD data on exit is not implemented yet, but it shouldn't be hard to do.

### DIFF
--- a/bochs/memory/memory-bochs.h
+++ b/bochs/memory/memory-bochs.h
@@ -148,6 +148,7 @@ private:
   Bit8u   flash_type;
   Bit8u   flash_status;
   Bit8u   flash_wsm_state;
+  bool    flash_modified;
 
   BX_MEM_SMF Bit8u flash_read(Bit32u addr);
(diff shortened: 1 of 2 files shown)


stlintel (approved):

fysnet:
Volker,

Thanks for adding this code. I can confirm that with PR #274, the ESCD is preserved when the `flash_data=` parameter is used.

I think the only thing I would add is if a filename is given that is not found, the file is created and cleared to all 0xFF's. Currently, if the file is not found, the values are not loaded, and even worse, the values are not saved, with no user notification that they were not.

Maybe if the filename is not found, a dialog is shown asking if the user would like to create it. A simple, "yes", "no", "cancel" dialog, and if the "cancel" is pressed, the emulation aborts.

What do you think?

Thank you,
Ben

PR #169, opened 2023-12-15
Fix memory corruption when reading Voodoo ROM

### LINKED ISSUE #131: Crash of Bochs / OS with default Voodoo 3 drivers
When I try to load Windows XP SP3 with default Voodoo 3 drivers, one of two things happens:
1. Either Bochs crashes with `>>PANIC<< APIC read at address 0x0000fee00081 spans 32-bit boundary !`;
2. Or OS crashes with `TSS selector points to bad TSS`.

This problem start to appear after updating from 620d0912646261c7e5b20fe69ef85823637abc85 to f5b54a4d336d84caf7eb744f6f97ae768c317f01.
Most likely, f5b54a4d336d84caf7eb744f6f97ae768c317f01 is not direct cause of the problem, but it uncovers problem somehow.

Vort:
Fixes #131.

### DIFF
--- a/bochs/iodev/display/banshee.cc
+++ b/bochs/iodev/display/banshee.cc
@@ -833,6 +833,8 @@ void bx_banshee_c::mem_read(bx_phy_address addr, unsigned len, void *data)
         case 1:
           *((Bit8u*)data) = (Bit8u)value;
           break;
+        case 3:
+          *((Bit8u*)data + 2) = (Bit8u)(value >> 16);
         case 2:
           *((Bit16u*)data) = (Bit16u)value;
           break;


Vort:
I took idea of how to handle `len` = `3` reads from
https://github.com/bochs-emu/Bochs/blob/50258bf32e5b4a7db7b20d4294ed5beea85b9bfe/bochs/iodev/usb/usb_ehci.cc#L896-L900
After further thought, doubts about whether this code will work correctly on big-endian architectures appeared.
However I have no access to such machines, so I can't test it.

stlintel (approved):

PR #197, opened 2023-12-28
Debugger always enabled

stlintel:
Please re-generate configure script before use :)

Allow to have debugger and non-debugger version in same binary.

Just compile Bochs with --enable-debugger and specify -debugger option in command line - and it run Bochs internal debugger. If -debugger not supplied it will run normal optimized build.

This is first draft of the idea provided for initial testing.

When -debugger is not supplied emulation speed is affected ~1% which I think if very reasonable taking into account extra capability.

When -debugger is supplied the emulation speed is affected severely as before.

There are potential ways how to make debugger cost cheaper (not addressed here yet):
- data watchpoints could be implemented in cheap way so they won't cost any emulation performance if not used (meaning there are no active watchpoints set).
- similarly code breakpoints could be optimized using the same mechanisms

Also:
- it would be nice to have GUI debugger button which will switch into debugger even if started as 'optimized'
 and if already done it would be nice to do 'optimized continue' which will switch to fast mode until some even drops you into the debugger back.

Also code is just hacky for now, will be improving the logic while doing more testing.
Especially need to test without handlers chaining compiled and when SMP is enabled.

For now I need as much feedback as possible.
What to improve ?
Bugs ?
How to make it more user friendly ?


bochs/cpu/cpu.cc
```diff
@@ -338,9 +415,7 @@ void BX_CPP_AttrRegparmN(2) BX_CPU_C::repeat(bxInstruction_c *i, BxRepIterationP
       }
       if (RCX == 0) return;
 
-#if BX_DEBUGGER == 0
-      if (BX_CPU_THIS_PTR async_event)
-#endif
+      if (BX_CPU_THIS_PTR async_event || bx_dbg.debugger_active)
```
stlintel, line 418: we basically not allow traces when bx_dbg.debugger_active is set because debugger doesn't work well with them (many things are checked in main loop and not in the instruction handlers).
probably the easiest way to stop trace everywhere will be use smth like STOP_TRACE bit in async_event
May be invent one called ASYNC_EVENT_IN_DEBUGGER, set it from cpu_loop_debugger and will automatically cause exist to main loop after every instruction due to async_event being set

bochs/main.cc
```diff
@@ -1032,73 +1037,78 @@ int bx_begin_simulation(int argc, char *argv[])
...
-  bx_dbg_main();
-#else
-#if BX_GDBSTUB
-  // If using gdbstub, it will take control and call
-  // bx_init_hardware() and cpu_loop()
-  if (bx_dbg.gdbstub_enabled) bx_gdbstub_init();
+  if (bx_dbg.debugger_active) {
+    // If using the debugger, it will take control and call
+    // bx_init_hardware() and cpu_loop()
+    bx_dbg_main();
```
stlintel, line 1043: this starts debugger cpu_loop

bochs/cpu/cpu.h
```diff
@@ -5500,7 +5503,7 @@ class bxInstruction_c;
 
 #define BX_NEXT_INSTR(i) {                             \
   BX_COMMIT_INSTRUCTION(i);                            \
-  if (BX_CPU_THIS_PTR async_event) return;             \
+  if (BX_CPU_THIS_PTR async_event || bx_dbg.debugger_active) return;             \
```
stlintel, line 5506: we basically not allow traces when bx_dbg.debugger_active is set because debugger doesn't work well with them (many things are checked in main loop and not in the instruction handlers).
probably the easiest way to stop trace everywhere will be use smth like STOP_TRACE bit in async_event
May be invent one called ASYNC_EVENT_IN_DEBUGGER, set it from cpu_loop_debugger and will automatically cause exist to main loop after every instruction due to async_event being set

Mainly I want all this to avoid the "|| bx_dbg.debugger_active" here which actually reduce emulation speed a bit (~1% measured) but also code will be much simpler

bochs/cpu/cpu.cc
```diff
@@ -49,6 +49,80 @@
...
+
+    for(;;) {
+      if (BX_CPU_THIS_PTR trace)
+        debug_disasm_instruction(BX_CPU_THIS_PTR prev_rip);
+
+      // want to allow changing of the instruction inside instrumentation callback
+      BX_INSTR_BEFORE_EXECUTION(BX_CPU_ID, i);
+      RIP += i->ilen();
+      BX_CPU_CALL_METHOD(i->execute1, (i)); // might iterate repeat instruction
+//    BX_CPU_THIS_PTR prev_rip = RIP; // commit new RIP
```
stlintel, line 101: remove comments

bochs/cpu/icache.cc
```diff
@@ -199,6 +220,8 @@ bxICacheEntry_c* BX_CPU_C::serveICacheMiss(Bit32u eipBiased, bx_phy_address pAdd
 
 bool BX_CPU_C::mergeTraces(bxICacheEntry_c *entry, bxInstruction_c *i, bx_phy_address pAddr)
 {
+  BX_ASSERT(!bx_dbg.debugger_active);
```
stlintel, line 223: mergeTraces should not be called with debugger active ever.
Also TODO: forming of traces is meaningless, may be we can just limit max length of a trace to 1 when debugger is running.

bochs/bx_debug/dbg_main.cc
```diff
@@ -702,6 +702,9 @@ void bx_dbg_lin_memory_access(unsigned cpu, bx_address lin, bx_phy_address phy,
 
 void bx_dbg_phy_memory_access(unsigned cpu, bx_phy_address phy, unsigned len, unsigned memtype, unsigned rw, unsigned access, Bit8u *data)
```
stlintel, line 703: the 

```
  if (! bx_dbg.debugger_active)
    return;
```

is here. This means function call is always done (will all the parameters) and it costs a lot.
Can be optimized significantly by extracting the 'if' into BX_DBG_LIN_MEMORY_ACCESS and BX_DBG_PHY_MEMORY_ACCESS macros

stlintel:
If someone interested in having this feature - would like to have feedback and discussion !

Vort:
First of all, I have no experience with Bochs debugger.
Also I'm used to graphical debugging UI, GDB-style looks convoluted for me.
I can try to run it anyway however.
But at this stage I can't even build it. Maybe my mistake, I don't know.
Here is what errors compiler produce:
```
ld.lld: error: undefined symbol: tputs
>>> referenced by libreadline.a(display.o):(rl_redisplay)
>>> referenced by libreadline.a(display.o):(rl_redisplay)
>>> referenced by libreadline.a(display.o):(rl_redisplay)
>>> referenced 19 more times

ld.lld: error: undefined symbol: tgetnum
>>> referenced by libreadline.a(terminal.o):(_rl_get_screen_size)
>>> referenced by libreadline.a(terminal.o):(_rl_get_screen_size)

ld.lld: error: undefined symbol: tgetent
>>> referenced by libreadline.a(terminal.o):(_rl_init_terminal_io)

ld.lld: error: undefined symbol: tgetstr
>>> referenced by libreadline.a(terminal.o):(_rl_init_terminal_io)

ld.lld: error: undefined symbol: tgetflag
>>> referenced by libreadline.a(terminal.o):(_rl_init_terminal_io)
>>> referenced by libreadline.a(terminal.o):(_rl_init_terminal_io)
>>> referenced by libreadline.a(terminal.o):(_rl_init_terminal_io)

ld.lld: error: undefined symbol: BC
>>> referenced by libreadline.a(terminal.o):(.refptr.BC)

ld.lld: error: undefined symbol: PC
>>> referenced by libreadline.a(terminal.o):(.refptr.PC)

ld.lld: error: undefined symbol: UP
>>> referenced by libreadline.a(terminal.o):(.refptr.UP)
g++: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Makefile:181: bochs] Error 1
```

stlintel:
It supposed to work also with graphic debugger in place.
Enable display_library: win32, options="gui_debug"
For compilation error you see try to :

// adds support for the GNU readline library in the debugger command
// prompt.
#define HAVE_LIBREADLINE 0
#define HAVE_READLINE_HISTORY_H 0

in config.h

Vort:
> #define HAVE_LIBREADLINE 0

This helped.

> Enable display_library: win32, options="gui_debug"

Ok, it works.
Strange that default font is not monospace, good that ability to change it exists.

I have some questions about it, but I suspect they may be offtopic in this PR.
For example, how to see what is located at addresses starting from `ss:[ebx]` now, before it is overwritten by `si`? Should I do some manual math for it?
bochs_debugger

> Bugs ?

I see strange change.
When I click Power off, even without `-debugger` option I see additional line:
image

Another stange (and probably not important) thing:
Do not remember what I did, but suspicious message appeared:
_(fatal() should never return, but it just did)_
image

stlintel:
>For example, how to see what bytes are located at addresses starting from ss:[ebx] now, before they are overwritten by si? >Should I do some manual math for it?

There is always GUI debugger command line with all its options.
You could do "x ss:ebx" and it will print you the value at ss:[ebx].
GUI debugger also has memory watch, you should be able to find it in the menus.

About the default font, the font was selected by person who originally submitted the GUI debugger code.
I bet we could change a default font if you have better suggestion :)

stlintel:
BTW, may be instead of having this .bochsrc option we should just fire debugger gui automatically if option -debugger was supplied. Or if -debugger=gui was supplied ...

Vort:
> I bet we could change a default font if you have better suggestion :)

I think answer about which font to choose mainly depends on state of support of Windows XP as host OS.
Is it completely broken or there are just no official builds for it?
  
Also question above can me made non-relevant in case of implementing checking for font existence.

Here is the list of popular fonts:
* Consolas - Windows Vista+
* Lucida Console - Windows 98+
* Courier New - Windows 3.1+
* Fixedsys - Windows 1+

Also Cascadia Code exists, but it is probably available only on Windows 11.

So I think it is better to select Consolas if it can be found and if not, then Lucida Console.

<details>
  <summary>Here are screenshots:</summary>

**Consolas:**
bd_consolas

**Lucida Console:**
bd_lucida_console

**Courier New:**
bd_courier_new

**Fixedsys:**
bd_fixedsys

Default:
bd_default

</details>

Vort:
> You could do "x ss:ebx" and it will print you the value at ss:[ebx].

I was thinking more about how to fill this section with corresponding bytes:
image

bochs/cpu/cpu.h
stlintel: done

bochs/cpu/cpu.cc
stlintel: done

Vort:
_Sorry for probably stupid questions._

Title states "Debugger always enabled", but it is required to "compile Bochs with --enable-debugger".
Why enabling it if it is always enabled?

More precisely, I'm interested in how this change will affect release binaries (and binaries made with GitHub actions).
Will this feature be enabled by default or it will require building from source like before?

stlintel:
> _Sorry for probably stupid questions._
> 
> Title states "Debugger always enabled", but it is required to "compile Bochs with --enable-debugger". Why enabling it if it is always enabled?
> 
> More precisely, I'm interested in how this change will affect release binaries (and binaries made with GitHub actions). Will this feature be enabled by default or it will require building from source like before?

The idea is to set --enable-debugger ON by default now and compile release binaries with debugger option ALWAYS. Meaning no separate debugger enabled binary anymore

Vort:
Will GUI debugger be also always enabled (by default)?

stlintel:
If supported in currently compiled gui - i guess yes

stlintel:
> Will GUI debugger be also always enabled (by default)?

Yes, if supported by currently present GUI

vruppert (commented):
I tried the new code with a Win95 guest system and Voodoo3 graphics emulation on Linux host. The 3D mode is a good check for the CPU performance since these early 3D cards need to do some parts of the 3D by software. When running the ValleyOfRa demo the IPS output in status bar shows > 60 M ips without debugger compiled in. With debugger compiled in, but disabled I cannot get more than 24 M ips. I have no stopwatch here to compare the boot time of the guest systems I can run here. I guess the performance loss is minimal for normal usage, but it's massive when running 3D stuff here. Maybe @Vort can run some tests with his guest systems and software to re-check this.

Vort:
I usually check performance with Half-Life rendering particular scene.
Here is how it looks with https://github.com/bochs-emu/Bochs/commit/b942d5caf564ab8db22ab2e1a4b22d875ed4dcca ([bochs-b942d5c.exe](https://github.com/bochs-emu/Bochs/actions/runs/11989971412/artifacts/2228455988)) and `ips=55000000`:
image
It works just barely slower than requested (53M IPS instead of 55M).

With https://github.com/Vort/Bochs/commit/b64d195832b9659f5ff1333c6b341714e968e0c1 ([bochs-b64d195.exe](https://github.com/Vort/Bochs/actions/runs/11995955303/artifacts/2229567431)) results are:
image
16M IPS instead of 53M IPS is not ~1% drop as described in first message, but more like ~300% drop.
Something is not right.

Vort:
> I guess the performance loss is minimal for normal usage

OS loading lags even more than 3D.
With `ips=70000000` and b942d5caf564ab8db22ab2e1a4b22d875ed4dcca, minimum IPS value during Windows 95 boot is 58M.
With https://github.com/Vort/Bochs/commit/b64d195832b9659f5ff1333c6b341714e968e0c1 and same config, minimum value is 11M.
This is ~500% drop.

vruppert:
Now I noticed another issue after adding '-debugger' to the command line. Depending on the gui setup Bochs enters the debugger command line or shows the debugger dialog box. After entering 'c' or pressing "Continue" button Bochs crashes while processing the BIOS code.

stlintel:
Crashes like segfault?

stlintel:
1% drop is expected when debugger is not enabled i.e. running bochs without -debugger option. When -debugger is supplied the code goes to completely different cpu_loop which can result in massive slowdown. Smth like 20-30% is perfectly acceptable. But yes, don't know about 300%
Did you try to use master branch Bochs compiled with debugger as a reference? It should as slow as new one with -debugger in command line.

vruppert:
Yes, it's a segfault. When I run Bochs under gdb control and use the command line debugger with "trace on" I can see that it segfaults at a normal command of the BIOS code. The last piece of code that gdb reports is cpu.cc line 104. I have not yet tried the new code on Windows.
When using the master branch code the ValleyOfRa demo runs at 65 M ips without and 40 M ips with debugger present. I'm always using .conf.linux shortcut script and run Bochs with the same bochsrc and disk image.

Vort:
> 1% drop is expected when debugger is not enabled i.e. running bochs without -debugger option.

This is how I was running it and getting 300-500% drop:
`d:\Bochs-2.8\bochs-b64d195.exe -q -f i:\__VirtualMachines\_Bochs\Windows95\bochsrc_voodoo3.bxrc`

I posted link to my branch which allowed to get binary with GitHub Actions.
Probably something is wrong there, but it is just merge and addition of `--enable-debugger --enable-debugger-gui`, so I don't think I was able to break something.

I did not tested `-debugger` yet, because performance drop without it should be solved first.

vruppert:
Update: Tests on Windows host compiled with VS2019
M ips: 57 (master) / 20 (debugger-always-enabled)
Debugger on: the same segfault as noted on Linux

stlintel:
> Update: Tests on Windows host compiled with VS2019
> M ips: 57 (master) / 20 (debugger-always-enabled)
> Debugger on: the same segfault as noted on Linux

What do you test, Volker?
Just windows boot ? What is the disk image ? I don't get the segfault here when testing Win11-Test ISO image ...

Vort:
No need to have disk image to see this crash.
It happens with empty project without images at all.

"Exception thrown at 0x0000000000000000 in bochs.exe: 0xC0000005: Access violation executing location 0x0000000000000000."

Stack trace:
```
>	bochs.exe!BX_CPU_C::MOV_EdIdR(bxInstruction_c * i) Line 40	C++
 	bochs.exe!BX_CPU_C::cpu_loop_debugger() Line 110	C++
 	bochs.exe!bx_dbg_continue_command(bool expression) Line 2167	C++
 	bochs.exe!bxparse() Line 517	C++
 	bochs.exe!bx_dbg_interpret_line(char * cmd) Line 318	C++
 	bochs.exe!bx_dbg_user_input_loop() Line 373	C++
 	bochs.exe!bx_dbg_main() Line 305	C++
 	bochs.exe!bx_begin_simulation(int argc, char * * argv) Line 1044	C++
 	bochs.exe!bx_real_sim_c::begin_simulation(int argc, char * * argv) Line 971	C++
 	bochs.exe!win32_ci_callback(void * userdata, ci_command_t command) Line 770	C++
 	bochs.exe!bx_real_sim_c::configuration_interface(const char * ignore, ci_command_t command) Line 939	C++
 	bochs.exe!bxmain() Line 336	C++
 	bochs.exe!main(int argc, char * * argv) Line 555	C++
```

vruppert:
Segfault always happens in the BIOS code before entering the boot process. The failing address seems to depend on the VGABIOS used.

stlintel:
Because BIOS is a ROM and doesn't change - the failure is 100% reproducible and 100% on the same place here

$ ./bochs -f ../../.bochsrc.win1124h2test -dbg
========================================================================
                     Bochs x86 Emulator 2.8.devel
             Built from GitHub snapshot after release 2.8
                  Compiled on Nov 24 2024 at 23:19:57
========================================================================
00000000000i[      ] BXSHARE not set. using system default '(null)'
00000000000i[      ] reading configuration from ../../.bochsrc.win1124h2test
00000000000i[      ] installing win32 module as the Bochs GUI
00000000000i[      ] using log file .bochs.log
Bochs internal debugger, type 'help' for help or 'c' to continue
Switching to CPU0
Next at t=0
(0) [0x0000fffffff0] f000:fff0 (unk. ctxt): jmpf 0xf000:e05b          ; ea5be000f0
<bochs:1> step 2138973
Next at t=2138973
(0) [0x0000000e01dc] 0010:00000000000e01dc (unk. ctxt): out dx, al                ; ee

This portout writes AL=0 to port DX=0xB2
It succeeds but seems like after all got broken

<bochs:3> step
Next at t=2138974
(0) [0x0000000e01dd] 0010:00000000000e01dd (unk. ctxt): mov edx, 0x000000b3       ; bab3000000
<bochs:4> step
Segmentation fault

The mov from 0xe01dd crash on fetch
linear page 0x00000000000e0000 maps to physical page 0x0000000e0000

stlintel:
OK, I have found and fixed segfault reason.
Let me know how I can reproduce extreme slowdown cases

Vort:
Slowdowns are related to sound, not 3D. Specifically, to Sound Blaster 16.
Debugger builds have no sb16 sound at all, but have lags instead of it.
I will try to get profiler report for this problem.

Vort:
I don't see anything related to sound here unfortunately:
image

So probably the best method of catching this problem is to try reproducing absence of sb16 sound in Windows 95 with this code.

stlintel:
> Slowdowns are related to sound, not 3D. Specifically, to Sound Blaster 16. Debugger builds have no sb16 sound at all, but have lags instead of it. I will try to get profiler report for this problem.

Make sense  this is why I might not see them. I don't even compile with sound enabled and don't bother enabling it in bochsrc too...

stlintel:
> With the usable debugger active mode I have a new test result with the ValleyOfRa demo: 33 M ips. So the debugger inactive mode is slower than the active one. Since the demo doesn't use sound, I assume that the bug is inside of the cpu code. There are a lot of BX_DEBUGGER cases and I guess that calling the uninitialized debugger could cause the slowdown.

Can you send disk image / instructions to reproduce?
Debugger inactive is slower than active is a nonsense and looks like a bug and want to try to debug

vruppert:
Since my test installation consists of a flat Win95 base image with an undoable image with the Voodoo3 stuff on and a VVFAT drive with a lot of testing stuff, I tried to set up an image for you. I know that you usually have no sound support present, so I turned the SB16 off...
Now I can say that sound really makes a difference, even if not active in the demo. With master branch I getting 66 M ips, new code with inactive debugger 60 M ips and debugger active 33 M ips. These values are acceptable, but we really need to review the ISA DMA stuff that is used by the SB16. If you have a simple Win95 disk image lying around I can send you the drivers and demo with instructions for testing.

BTW: When setting up Win95 to reboot, Bochs prints a debugger disassembly line and stops responding. This happens in debugger inactive mode and the last log entries are reset messages from the devices.

vruppert:
I have now sent you some testing stuff with an e-mail. Please see the README.txt what to do with it. In the meantime I have tested the ValleyOfRa demo with Win98SE guest and ES1370 soundcard. There no big slowdown (60 instead of 65 M ips), so I guess the SB16 and the ISA DMA stuff causing the strange slowdown.

vruppert:
[voodoo3.zip](https://github.com/user-attachments/files/17907522/voodoo3.zip)
Sending e-mail failed - please try this one.

vruppert:
I think that I have found the reason for the slowdown. The DMA handling in event.cc needs to be fixed. Adding bx_guard.async.dma = 1 in main.cc makes the sound work normally and the slowdown in the 3D demo is minimal.

stlintel:
> I think that I have found the reason for the slowdown. The DMA handling in event.cc needs to be fixed. Adding bx_guard.async.dma = 1 in main.cc makes the sound work normally and the slowdown in the 3D demo is minimal.

I was also guessing the same. But I am not expert of what bx_guard.async.dma does and why it is needed to be 1.
This code existed even before I joined to Bochs ... and it was long time ago

stlintel:
I made some simple attempt to fix it, can you try it now ?

Vort:
I checked [bochs-0e51dc2.exe](https://github.com/bochs-emu/Bochs/actions/runs/12018587418/artifacts/2235396083) binary and results are good:
With Half-Life test, performance drop is only 53 M -> 51 M IPS.
However, my computer is under load, so this result is not precise.
For completeness, same test with `-dbg` shows 24 M IPS, which is good as well:
When debugging, performance is not that critical.

vruppert:
The test with the latest code is okay. The next thing to fix is the OS reboot issue (debugger inactive). At this point Bochs is no longer reponding:

00000000000i[      ] installing x module as the Bochs GUI
00000000000i[      ] using log file bochsout.txt
(0).[8352218428] [0x0000c0001250] 0028:00000000c0001250 (unk. ctxt): inc dword ptr ds:[ecx]    ; ff01

This is shown in terminal / console and the last messages in are from devices reset.

vruppert:
In line 1007 of exception.cc it is required to check whether or not the debugger is active. The ugly disasm out is still there, but Bochs continues working after this change. Maybe removing BX_DEBUGGER sections from normal cpu_loop() can fix this, too.

stlintel:
> disassembly

I have cleaned up more stuff following after BX_DEBUGGER ifdefs in the code. 

A few more left that I didn't yet know how to handle:

in main.cc:
```
  BX_DEBUG(("bx_init_hardware is setting signal handlers"));
// if not using debugger, then we can take control of SIGINT.
#if !BX_DEBUGGER
  signal(SIGINT, bx_signal_handler);
#endif
``` 

Why the signal restore is only when BX_DEBUGGER==0 ?

And another one in bx_atexit:

```
#if BX_DEBUGGER == 0
  if (SIM && SIM->get_init_done()) {
    for (int cpu=0; cpu<BX_SMP_PROCESSORS; cpu++)
#if BX_SUPPORT_SMP
      if (BX_CPU(cpu))
#endif
        BX_CPU(cpu)->atexit();
  }
#endif
```

Here I thin also worth calling BX_CPU->atexit() always regardless of BX_DEBUGGER and inside atexit() we will decide

What do you think ?

stlintel:
> With latest commits, Windows 95 reboots normally. Problem with unneeded disassembly lines still persists however. (But it is not of high priority I think)
> 
> ```
> (0).[2518368328] [0x000000148658] 0048:0000000000000258 (unk. ctxt): push ebp
>                ; 6655
> (0).[4056390484] [0x000000148658] 0048:0000000000000258 (unk. ctxt): push ebp
>                ; 6655
> ```

The disasm print should go away now

Vort:
I confirm that extra disasm lines during reboot and shutdown are gone.

stlintel:
The above two were resolved but this one is still a puzzle to me:

```
      if (usb_debug_dialog(type, param1, param2) < 0) {
        bx_user_quit = 1;
#if !BX_DEBUGGER
        bx_atexit();
        SIM->quit_sim(1);
#else
        bx_dbg_exit(1);
#endif
        return -1;
      }
```

Use bx_dbg.debugger_active here as well ?

fysnet:
> Shouldn't we better have single method for calling exit which will figure out what about BX_DEBUGGER and etc? Today these 6 lines of code are replicated accross many places ...

As you state, this is replicated across many places and I simply copy/pasted. I believe it would be better to have a single `exit` routine.

vruppert:
I think moving this duplicated config interface code for simulation shutdown to one function with correct debugger handling is required before this code can go to master branch. I haven't noticed other issues with it yet.

stlintel:
> I think moving this duplicated config interface code for simulation shutdown to one function with correct debugger handling is required before this code can go to master branch. I haven't noticed other issues with it yet.

I made necessary changes in prev commit.
Proper code would take this code into a function :

```
#if BX_DEBUGGER
        if (bx_dbg.debugger_active) {
          bx_dbg_exit(1);
        }
        else
#endif
        {
          bx_atexit();
          SIM->quit_sim(1);
        }
``` 

smth like:

```
void bx_quit(int exit_code)
{
#if BX_DEBUGGER
        if (bx_dbg.debugger_active) {
          bx_dbg_exit(exit_code);
        }
        else
#endif
        {
          bx_atexit();
          SIM->quit_sim(exit_code);
        }
}
```

donno if bx_quit is the best function name

vruppert:
I think the new code is now ready for the master branch. If issues appear, can fix them there.

vruppert (approved):

PR #198, opened 2023-12-29
AVX512-FP16 implementation

stlintel:
All softfloat methods for float16 are still missing -> I am temporarily using conversion to float32, computation as float32 and conversion back. In parallel investigating the idea of replacing the softfloat model in Bochs with softfloat3e.
Untested (unit tests are WIP)


stlintel:
Replace with new softfloat based model

PR #209, opened 2024-01-07
Allow the boot usb option

fysnet:
This allows the `usb` option to be used in the `boot:` parameter. Of course there is not support for booting from a USB, but this allows for a (probably much) further down the road addition.

This will break the current bios when booting from a network. However, the bios is already broken in this manner, especially when `BX_ELTORITO_BOOT` is not defined. The current bios code assumes and hard codes the BEV entry to 4. However, if the `BX_ELTORITO_BOOT` define is missing, the BEV will be added to the third entry, not the fourth entry.

For the BIOS to be updated correctly, things like the following will need to be updated instead of hard coded to `[4]`.
https://github.com/bochs-emu/Bochs/blob/54961f0326f9471931bc79e2734b9fa0e3c36c8d/bochs/bios/rombios.c#L2205

The following will need to be not hard coded to 4.
https://github.com/bochs-emu/Bochs/blob/54961f0326f9471931bc79e2734b9fa0e3c36c8d/bochs/bios/rombios.c#L2253-L2257

The following will need "usb" added.
https://github.com/bochs-emu/Bochs/blob/54961f0326f9471931bc79e2734b9fa0e3c36c8d/bochs/bios/rombios.c#L2102

The following assumes the next entry will be the fourth entry, not checking to see if `BX_ELTORITO_BOOT` was defined.
https://github.com/bochs-emu/Bochs/blob/54961f0326f9471931bc79e2734b9fa0e3c36c8d/bochs/bios/rombios.c#L10851-L10857

The following comment will need to be updated.
https://github.com/bochs-emu/Bochs/blob/54961f0326f9471931bc79e2734b9fa0e3c36c8d/bochs/bios/rombios.c#L8365-L8371

As well as a few other fixes and updates.

This PR simply allows the value of '4' (`BX_BOOT_USB`) to be placed in the boot sequence options, moving the network (`BX_BOOT_NETWORK`) to the next number.

The boot sequence option still only allows three parameters.
`boot: cdrom, floppy, disk`
or
`boot: cdrom, usb, disk`
I don't think it is necessary to add a fourth.


vruppert:
Having support for booting from USB in Bochs would be nice, but I think it doesn't make sense to apply these changes to master until there is BIOS support for it. With SeaBIOS booting from USB should work, but I never tried it and I don't know how it is implemented. This pull request should be postponed until after release 2.8. Otherwise users may try the new config choice and reporting issues here.

fysnet:
> Having support for booting from USB in Bochs would be nice, but I think it doesn't make sense to apply these changes to master until there is BIOS support for it.

Agreed. However, BIOS support is currently a work in progress, though it may be a little while. Also, I have been told, and agree to, that small frequent PRs are much more advisable than a single large PR, hence this PR.

> With SeaBIOS booting from USB should work, but I never tried it and I don't know how it is implemented. 

Without mandatory user intervention, how do you tell SeaBIOS to boot the USB? Unless I am mistaken, there is no current way in Bochs to tell SeaBIOS or any BIOS for that matter, to boot the USB (without mandatory user intervention, which requires the user to know how to intervene).

> This pull request should be postponed until after release 2.8.

Agreed. I have no problem with this.

> Otherwise users may try the new config choice and reporting issues here.

If it is not included in the documentation yet (.bochsrc, user.dbk, etc.), only those that read and keep up with these PRs would know to try it, and in doing so, would know that it isn't available yet.

Anyway, just my opinion. It is more difficult for me to maintain and have a code base that is not in sync with the Bochs master, so I was hoping that small simple PRs would be accepted. It makes it easier on both me and the Bochs master to keep track of these small changes rather than one large change.

However, do as you feel best. I will manage.

Thanks.

fysnet:
Just curious if this PR will ever be merged. I have the source to the referenced BIOS ready to post somewhere, I am simply waiting for the ability to boot from USB, the initial reason I wrote the BIOS.  Just curious.

vruppert:
The USB boot option only makes sense with a BIOS that supports it. I suggest to add the new BIOS sources as new repository inside of "bochs-emu" just like I did it with the LGPL'd VGABIOS. Then we could add the binary in the bochs/bios folder. You may also have a look at the current status of the USB debugger. I made it work with plugins support and started porting it to GTK.

fysnet:
The source code to my BIOS is at https://github.com/fysnet/i440fx. I think it will be easier for me to maintain there since I have write permission there and not here. Feel free to include the `i440fx.bin` file in Bochs' BIOS folder.  As always, I appreciate any feedback and improvements.

vruppert:
Since it's on Github, it should be okay for now. I just tried to compile the BIOS here, but since I'm on Linux, I'm using nbasm with wine. It fails to compile, but I don't know whether or not it's a wine issue. I can open an issue on your page, so can look for a workaround. Now it would be possible to apply the USB boot option, but I don't know whether or not the new BIOS can be included in the Bochs tree. The LGPL license has no restrictions for the type of usage, but if I understand correctly, nbasm and software created with it are for non-commercial usage only.

bochs/gui/siminterface.h
vruppert, line 1: It is necessary to remove the unused BX_BOOT_PCMCIA entry. Otherwise "boot: network" causes a panic. I have now managed to modify the Bochs BIOS and boot from network. Since I'm not familiar with the boot code, I had to add a dummy USB entry in addition to your BIOS hints.

fysnet: Looking into this a little more closely, it looks like I may need to modify the list slightly. In the Bios Boot Specification v1.01, page 17 has a list. It lists a zero based entry of 3 and 4 as BEV entries.
```
0 Floppy A:
1 Hard Drive C:
2 CD-ROM
3 BEV #1
4 BEV #2
```
However, on page 32, it lists the `type` of entries and has the network as entry 6.
```
00h = Reserved
01h = Floppy
02h = Hard disk
03h = CD-ROM
04h = PCMCIA
05h = USB device
06h = Embedded network
07h..7Fh = Reserved
80h = BEV device
81h..FEh = Reserved
```
I will need to look into this a little more.

vruppert: In the Bochs code many things related to the boot devices are index-based. In the config interface we need an empty string for the PCMCIA entry and textxonfig / win32paramdialog need to skip it when building the selection item. The BIOS code also needs to use e->type instead of the index.

vruppert (approved):

PR #224, opened 2024-01-15
Cancel topmost mode after leaving fullscreen mode

Vort:
`HWND_TOPMOST` should be cancelled with `HWND_NOTOPMOST` when it is not needed anymore.


vruppert:
What should this change do? I didn't notice a visible difference. I expected restoring the window position, but that's not the case.

Vort:
Bochs window should not be forced on top of other windows after leaving fullscreen mode.
This is what was happening here on Windows 7 host before this change.

vruppert:
Okay, since my Bochs window was on top before fullscreen mode I didn't see a difference afterwards.

vruppert (approved):

PR #277, opened 2024-03-10
Add static link support for bxhub and bximage

Vort:


stlintel:
Hi, can you take a look why cross compilation Ci/DI you added to build Win32 binary unable to link softfloat3e ?
Regular Ci/Di compilation works well...

Vort:
@stlintel because `bochs\cpu\softfloat3e\Makefile` should honor `CXX` option set by `.conf.win64-cross-mingw32`.
Other options like optimizations and debug information may make sense to propagate as well.

https://github.com/bochs-emu/Bochs/blob/f31b115e20e0b1ad3844e2d5d5f22c7e6cb1eb02/bochs/cpu/softfloat3e/Makefile#L45
https://github.com/bochs-emu/Bochs/blob/f31b115e20e0b1ad3844e2d5d5f22c7e6cb1eb02/bochs/.conf.win64-cross-mingw32#L7

Vort:
Here is my suggestion to fixing this problem: #279.

PR #284, opened 2024-03-14
Bump `BIOSROMSZ` to 4M to accommodate OVMF

DrChat:
OVMF is 4 megabytes by default, so in order to avoid customizing their build process we'll want to accommodate a 4M BIOS in Bochs.

### DIFF
--- a/bochs/memory/memory-bochs.h
+++ b/bochs/memory/memory-bochs.h
@@ -36,7 +36,7 @@
 
 class BX_CPU_C;
                                        // 512K BIOS ROM @0xfff80000
-const Bit32u BIOSROMSZ = (1 << 21);    //   2M BIOS ROM @0xffe00000, must be a power of 2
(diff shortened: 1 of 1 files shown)


stlintel:
Ben, Volker - agree ?

fysnet:
> Ben, Volker - agree ?

I have not done a lot of UEFI yet, just enough to get familiar with it and how it works. So I am not sure, but I do remember reading that it does require 4-meg, so I agree, I believe this PR to be a valid request.
Ben

DrChat:
If you want to play around with UEFI in action - you can build [OVMF](https://github.com/tianocore/edk2) based off of tag `edk2-stable202105` with this patchset:
[bochs.patch](https://github.com/bochs-emu/Bochs/files/14618999/bochs.patch)

Boots right up :)

vruppert (approved):

PR #297, opened 2024-04-01
telescope for debugger

therealdreg:
**telescope** aims to simplify the dereferencing of an address to determine the content it actually points to. At a glance, you can see the entire chain + final data (ascii string / hex). Very handy for reversing & exploiting.

``` 
  telescope addr               shows how pointers lead to other pointers and finally to data. 
                               It goes step by step through each pointer, showing you where they all end up. 
                               To activate this functionality, the telescopeset command must be executed beforehand. 
                               For example: telescope esp
                               it presents a sequential view where each row represents the data at the next address 
                               starting from the original location. This means that the command not only shows the 
                               chain of pointers and the end data but also arranges the output so that each row 
                               progresses to the next memory address from the starting point
        
  telescopeset max_rows max_depth data_depth data_hex
                               The number of rows ('max_rows') to display, how deeply to follow pointer dereferencing 
                               ('max_depth'), the depth of final displayed data ('data_depth'), and the option to 
                               show the final data in hexadecimal(1)/ascii(0) format flag ('data_hex'). 
                               This command also activates the telescope feature for viewing registers and the stack. 
                               For example: telescopeset 6 5 15 0
                               display up to 6 addresses, dereference pointers 5 levels deep, show 15 units of data in ASCII
                               (0 means not use hexadecimal)
                               Disable telescope: telescopeset 0 0 0 0
```
Example: **telescopeset 6 5 15 0**

Once activated with the **telescopeset** command, it can be viewed with the **regs**, **print-stack**, and used individually with the **telescope** command:

image

- It also detects infinite loops (**iloop**)
- The final data can be displayed in hexadecimal or ascii-string (filtered)
- **telescope** command
- disable telescope: **telescopeset 0 0 0 0**

image

A wild PE32 appeared!

image

image

----

I would like the first version of this feature to be **simple** and straightforward, **without adding many options for tuning**.

Since the current code is tightly coupled with the UI, I've made the minimum necessary changes to make it functional and "non-noisy"

Do you like it? What changes would you like me to make? 

**I haven't fixed the coding style yet, and it's not completely finished, but I wanted to discuss the idea before continuing!**

@stlintel @vruppert :D

----

Inspired by GDB-GEF, PEDA, PWNDBG...

GDB-GEF:

image

PWNDBG:

image

-----


therealdreg:
I think the workflow fails because I'm modifying the parser & lexer :?. If I do it in the correct order on my machine, everything compiles fine...

stlintel:
You clearly have to write better doc about what is telescope and how it is configured

telescopeset max_rows max_depth data_depth data_hex

There max_rows max_depth data_depth are just magic numbers from first sight

also I wonder where the name comes from

therealdreg:
> You clearly have to write better doc about what is telescope and how it is configured
> 
> telescopeset max_rows max_depth data_depth data_hex
> 
> There max_rows max_depth data_depth are just magic numbers from first sight
> 
> also I wonder where the name comes from

Yes, its not the final version, but I want more feedback.

About name 'telescope':

https://browserpwndbg.readthedocs.io/en/docs/commands/procinfo/telescope/
https://github.com/ebtaleb/peda_cheatsheet/blob/master/peda.md#stack--memory
https://gef-legacy.readthedocs.io/en/latest/commands/dereference/

therealdreg:
@stlintel Besides improving the documentation, fixing the coding style, enhancing the code... is there anything else that comes to mind?

therealdreg:
fixed coding style & doc @stlintel opinion? (my english s*cks)

bochs/gui/enh_dbg.cc
```diff
@@ -523,7 +523,7 @@ void UpdateStatus()
 // Note: laddr + len must not cross a 4K boundary -- otherwise, there are no limits
 bool ReadBxLMem(Bit64u laddr, unsigned len, Bit8u *buf)
 {
-    return bx_dbg_read_linear(CurrentCPU, laddr, len, buf);
+    return bx_dbg_read_linear(CurrentCPU, laddr, len, buf, false);
```
stlintel, line 526: please make the param 'quiet' as default false and void changing every call to bx_dbg_read_linear

bochs/bx_debug/lexer.l
```diff
@@ -117,6 +117,8 @@ writemem        { bxlval.sval = strdup(bxtext); return(BX_TOKEN_WRITEMEM); }
 loadmem         { bxlval.sval = strdup(bxtext); return(BX_TOKEN_LOADMEM); }
 setpmem         { bxlval.sval = strdup(bxtext); return(BX_TOKEN_SETPMEM); }
 deref           { bxlval.sval = strdup(bxtext); return(BX_TOKEN_DEREF); }
+telescope       { bxlval.sval = strdup(bxtext); return(BX_TOKEN_TELESCOPE); }
+telescopeset    { bxlval.sval = strdup(bxtext); return(BX_TOKEN_TELESCOPESET); }
```
stlintel, line 121: might be also telescope reset shortcut instead of 'telescope 0 0 0 0' ?
also lexer is not forgiving for these 4 params
if you forget one or miss order of these params -> result is not nice and hard to understand what went wrong
I prefer smth like telescope rows=R depth=D data_depth=DD hex=H
but I didn't conclude it yet ... anyway -> at least when telescope is called -> print what is actually did.
like setting 'rows = R and etc
Are where default params here ?
Like hex=TRUE is omitted ?
Or data_depth = max_depth if omitted ?

bochs/bx_debug/debug.h
```diff
@@ -80,7 +80,7 @@ void bx_dbg_set_magic_bp_mask(Bit8u new_mask);
 void bx_dbg_clr_magic_bp_mask(Bit8u mask);
 Bit8u bx_dbg_get_magic_bp_mask_from_str(const char *str);
 void bx_dbg_print_magic_bp_mask_from_str(Bit8u mask);
-bool bx_dbg_read_linear(unsigned which_cpu, bx_address laddr, unsigned len, Bit8u *buf);
+bool bx_dbg_read_linear(unsigned which_cpu, bx_address laddr, unsigned len, Bit8u *buf, bool quiet);
```
stlintel, line 83: use = false default parameter

bochs/bx_debug/dbg_main.cc
```diff
@@ -1245,56 +1251,96 @@ void bx_dbg_info_registers_command(unsigned cpu, int which_regs_mask)
 #endif
 #if BX_SUPPORT_X86_64 == 0
     reg = BX_CPU(cpu)->get_reg32(BX_32BIT_REG_EAX);
-    dbg_printf("eax: 0x%08x %d\n", (unsigned) reg, (int) reg);
+    dbg_printf("eax: 0x%08x %d ", (unsigned) reg, (int) reg);
```
stlintel, line 1254: awful code duplication.
should be array of const char* containing "rax", "rbx", ... , "r15" like this one from disasm.cc:

static const char *general_64bit_regname[17] = {
    "rax", "rcx", "rdx", "rbx", "rsp", "rbp", "rsi", "rdi",
    "r8",  "r9",  "r10", "r11", "r12", "r13", "r14", "r15", "rip"
};

and for walking over all registers

bochs/bx_debug/dbg_main.cc
```diff
@@ -1939,6 +1987,119 @@ void bx_dbg_deref_command(bx_address addr, unsigned deep)
...
+      dbg_printf("0x%08x%08x ", GET32H(addr), GET32L(addr));
+      bx_dbg_telescope(addr);
+      dbg_printf("\n");
+    }
+    else
+#endif
+    {
+      if (len == 4) {
+        dbg_printf("0x%08x ", (unsigned) addr);
+        bx_dbg_telescope(addr);
```
stlintel, line 2089: Avoid duplicating bx_dbg_telescope(addr) call ?
just do dbg_printf with if's and call bx_dbg_telescope(addr) after them

bochs/bx_debug/parser.y
```diff
@@ -969,6 +973,22 @@ deref_command:
...
+telescope_command:
+      BX_TOKEN_TELESCOPE expression '\n'
+      {
+        bx_dbg_telescope_command($2);
+        free($1);
+      }
+    ;
+
+telescopeset_command:
+      BX_TOKEN_TELESCOPESET BX_TOKEN_NUMERIC BX_TOKEN_NUMERIC BX_TOKEN_NUMERIC BX_TOKEN_NUMERIC '\n'
```
stlintel, line 985: is possible to have default parameters ?

bochs/bx_debug/parser.y
```diff
@@ -1545,7 +1575,7 @@ expression:
    | expression '-' expression       { $$ = $1 - $3; }
    | expression '*' expression       { $$ = $1 * $3; }
    | expression '/' expression       { $$ = ($3 != 0) ? $1 / $3 : 0; }
-   | expression BX_TOKEN_DEREF_CHR expression { $$ = bx_dbg_deref($1, $3, NULL, NULL); }
+   | expression BX_TOKEN_DEREF_CHR expression { $$ = bx_dbg_deref($1, $3, NULL, NULL, 0); }
    | expression BX_TOKEN_RSHIFT expression { $$ = $1 >> $3; }
```
stlintel, line 1579: I would probably prefer default parameter also here for bx_dbg_deref

stlintel (changes requested):

stlintel:
Please also include generated lex and yacc files in the PR

sskras:
@therealdreg commented [on Apr 1, 2024](https://github.com/bochs-emu/Bochs/pull/297#issuecomment-2029823103):

> About name 'telescope':
> 
> **_https://browserpwndbg.readthedocs.io/en/docs/commands/procinfo/telescope/_**
> https://github.com/ebtaleb/peda_cheatsheet/blob/master/peda.md#stack--memory
> https://gef-legacy.readthedocs.io/en/latest/commands/dereference/

The first URLs is broken now, the root URL says:

> > #### 404 Project not found
> > [<sup>browserpwndbg.readthedocs.io</sup>](https://browserpwndbg.readthedocs.io/)
> > The project you requested does not exist or may have been removed.

PR #307, opened 2024-04-16
Prevent a serial RX overflow if the port is in FIFO mode

DrChat:
1) If the guest's serial port is in FIFO mode and is almost full, throttle new inputs so that an RX overflow error isn't triggered. This is necessary because a host process on the other end is much faster than the guest, so it can easily overflow the port and cause the guest to discard data.
1) Open named pipes in an asynchronous mode, just so something is not required to connect to the pipe on startup. This has been a useful change for me, but if it is too problematic I can remove it.

### DIFF
--- a/bochs/iodev/serial.cc
+++ b/bochs/iodev/serial.cc
@@ -544,8 +544,8 @@ bx_serial_c::init(void)
           if (server) {
             // server mode
             pipe = CreateNamedPipe( dev,
-                PIPE_ACCESS_DUPLEX | FILE_FLAG_FIRST_PIPE_INSTANCE,
-                PIPE_TYPE_BYTE | PIPE_READMODE_BYTE | PIPE_WAIT,
+                PIPE_ACCESS_DUPLEX,
+                PIPE_TYPE_BYTE | PIPE_READMODE_BYTE | PIPE_NOWAIT,
                 1, 4096, 4096, 0, NULL);
 
             if (pipe == INVALID_HANDLE_VALUE) {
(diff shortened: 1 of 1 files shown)


bochs/iodev/serial.cc
```diff
@@ -1538,8 +1539,10 @@ void bx_serial_c::rx_timer(void)
     if (BX_SER_THIS s[port].tty_id >= 0) FD_SET(BX_SER_THIS s[port].tty_id, &fds);
 #endif
   }
+
+  // N.B: Throttle inputs if the OS has not yet responded to us.
   if ((BX_SER_THIS s[port].line_status.rxdata_ready == 0) ||
-      (BX_SER_THIS s[port].fifo_cntl.enable)) {
+      ((BX_SER_THIS s[port].fifo_cntl.enable) && (BX_SER_THIS s[port].rx_fifo_end < 15))) {
```
DrChat, line 1545: Not sure what your thoughts are on applying this to `BX_SER_MODE_RAW` - that's the only odd one out here.

vruppert (commented):
I never tried the pipe-server / pipe-client mode, so I can only confirm that this code compiles without issues.

DrChat:
@vruppert I've confirmed that it all works locally, though I cannot confirm if it breaks anyone's expectations.
I'm happy to revert that part of this change if no-one is able to confirm that the change is okay.

stlintel:
is this PR still relevant or already should be dropped ?

vruppert:
I haven't worked with serial port for a long time and I never tried the "pipe" modes. At the moment I have no chance to test it.

DrChat:
The fix on line 1545 is definitely still relevant. Let me follow up and drop the other changes :)

PR #378, opened 2024-11-04
Get DESTDIR from the environment and add distclean target.

sortie:
autoconf-based packages conventionally respect the DESTDIR environment variable rather needing to explicitly override it.

Likewise the distclean target is conventionally named distclean rather than dist-clean. This change adds a top-level alias to be compatible.

These changes are needed to build bochs on my infrastructure without having to special case bochs or locally patch it.


stlintel (approved):

sortie:
Wow fast -- thanks! :)

PR #390, opened 2024-11-17
remove generic cpuid model and all releavnt params

stlintel:
introduce i387 cpuid to be to configure minimalistic i386 cpu without recompiling with BX_CPU_LEVEL=3 all other possible things could be configured using add_feature, exclude_feature knobs


stlintel:
testing required: 
1. the CPUID menu should be removed successfully from all GUIs
2. it should be possible to configure i386 CPUID even if not compiled with CPU-LEVEL=3
   but in fact I was not able to do so. the i386 cannot run with latest Bochs BIOS and require Legacy BIOS.

Ideally one need to run latest full featured Bochs with 386 configuration and legacy BIOS, boot to MSDOS and run some CPUID tools to see this is indeed i386. It is also possible to use "disable_feature" and disable "fpu" to see 386 without FPU support.
Everything else (all other CPUID models) supposed to work as usual and not be affected.

stlintel:
[Cpuutil.zip](https://github.com/user-attachments/files/17791812/Cpuutil.zip)

I don't know but might be utils from there could recognize also i386 and tell it is 386

stlintel:
Update: managed to find DOS6.22 image and CHKCPU utility to identify 386
http://www.steunebrink.info/chkcpu.htm

Here is the result:
image

Seems like all working

Vort:
I like these changes overall, but I doubt `model=i386` is good default option.
I think CPU, which provides good amount of acceleration, but also capable of running most OSes will fit better.
(`core2_penryn_t9600` maybe? it is already mentioned as example in `bochs\.bochsrc`)

> I also can confirm Win32 GUI has no CPUID option anymore

Same for wxWidgets
image

Vort:
Also default CPU model should be compatible with default BIOS.

stlintel:
The default is selected while creating param:

```
  new bx_param_enum_c(cpu_param,
      "model", "CPU configuration",
      "Choose pre-defined CPU configuration",
      cpu_names, 0, 0);
```
The first 0 is the initial value.
In order to set any meaningful value need to return proper number matching corresponding CPU model
It is possible to create a function which will return preferable CPU model for every BX_CPU_LEVEL
Like:

```
#if BX_CPU_LEVEL== 4
return bx_cpudb_model_i486dx4;
#elif  BX_CPU_LEVEL== 5
return bx_cpudb_model_pentium_mmx;
#elif BX_CPU_LEVEL==6
  #if BX_SUPPORT_X86_64
   return ...
```
and etc

The problem that enum is defined only in cpu/init.cc and not really available in config.cc
From the other hand the function could be also in init.cc
But I think changing default could be next step.

stlintel:
> Also default CPU model should be compatible with default BIOS.

Pushed an update, what do you think now ?

Vort:
>Pushed an update, what do you think now ?

It means `corei7_haswell_4770` for default builds.
I checked it with DOS 6.22, Windows 3.11, 95 and 98 - all work fine.

vruppert:
I can confirm that the textconfig changes are okay. I also had a look at the add/remove features function and noticed that there is no help message for the available features. For the gui dialogs we'll need a solution to configure comma separated lists. I think this could be done similar to the optional plugin control.

Issue #393, opened 2024-11-21
Incomplete page table in GUI debugger


Vort:
I needed to look at page table near address `0xCAD00000`, but GUI debugger wasn't able to show it correctly:
![image](https://github.com/user-attachments/assets/d07880b5-2cfc-4b18-8eb8-e545d8c0aa90)

Luckily, I was able to get this information with `info tab` text command.
However, its output is incomplete as well, it can't be scrolled to addresses before `0xCA3DD000`.

Version: dbfb11f1

Vort:
Turned out, reproduction is not as easy as I thought.
On one installation of Windows 95 table shows correctly, on other - not correctly.

However, I think problem appear just when page table gets too large, `LineCount < 1024` looks suspicious here:
https://github.com/bochs-emu/Bochs/blob/dbfb11f1ddd42f784af5089e24ec16cfaf683555/bochs/gui/enh_dbg.cc#L1646
I will think how to force OSes to generate large page tables.

Vort:
Not the smartest piece of code, but whatever: [memalloc.zip](https://github.com/user-attachments/files/17848432/memalloc.zip).
Launch 2-4 instances of it simultaneously in Windows 95 and page table will grow enough for problem reproduction.

Vort:
By the way, to update page table in GUI after Continue - Break cycle, Alt+F2 should be pressed twice.
I consider this as yet another bug. GUI table should update automatically.

stlintel:
> By the way, to update page table in GUI after Continue - Break cycle, Alt+F2 should be pressed twice.
> I consider this as yet another bug. GUI table should update automatically.

In order to update the table efficiently need to catch if it was modified since last update, which is nearly impossible.
BTW, the table itself isn't really trustful too because CPU has TLB and TLB is not necessary coherent with memory.

Vort:
In cases when autoupdate is not desirable, GUIs usually provide "Refresh" button, which helps user to understand that table is not autoupdated.

If autoupdate is that problematic, then it is possible to focus this issue specifically on artificial table length limit. I checked and confirmed that `LineCount < 1024` is the cause.

stlintel:
I noticed that GUI debugger dump page table routine was kind of outdate, text debugger used different routine which was updated to support 64-bit and more.
I ported it to GUI debugger but haven't test it yet - can you take a look and if okay, will merge it

stlintel:
Let me know if the issue may be considered to be resolved
[...]

PR #395, opened 2024-11-23
port GUI page table dump routine from text debugger to GUI debugger code

stlintel:
it is copy-paste but I didn't find a way to do it better yet

supposed to help with https://github.com/bochs-emu/Bochs/issues/393

### DIFF
--- a/bochs/gui/enh_dbg.cc
+++ b/bochs/gui/enh_dbg.cc
@@ -1631,28 +1631,27 @@ void AddPagingLine(int LC, char *pa_lin, char *pa_phy)
 // lifted from bx_dbg_dump_table in dbg_main of the internal debugger
 void FillPAGE()
 {
-    Bit32u lin, start_lin, curlin; // show only low 32 bit
-    bx_phy_address phy;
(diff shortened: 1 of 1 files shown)


Vort:
1. [I checked and confirmed that LineCount < 1024 is the cause.](https://github.com/bochs-emu/Bochs/issues/393#issuecomment-2495482218) [Your code still contains this check.](https://github.com/bochs-emu/Bochs/blob/335cb11078165d136afef2767114c7c5a1a13b86/bochs/gui/enh_dbg.cc#L1650) So problem from #393 remains in place.
2. `bool valid;` is missing.

stlintel:
Without Line <= 1024 it might print too many pages and printing too many pages might take too much time.
Is it safe to let him traversal entire page table which might contains hundreds of thousands pages in worst case ?

stlintel:
Make it 10000 would be good for your most of the needs, what do you say ?

Vort:
> Make it 10000 would be good for your most of the needs, what do you say ?

Ok.

Proper solution is to split page table into... pages.
But that's a task for another time.

PR #415, opened 2024-12-15
UHCI: Fix erroneous frame number wrapping behaviour

Cacodemon345:
This PR fixes the frame number wrapping behaviour.

Previously it would incorrectly wrap around 1023 (0x3FF) instead of 2047 (0x7FF), causing behaviour to not match the specification. This PR makes it wrap around correctly.


fysnet (approved):
Yes, this is correct. I don't know how I missed it. Thank you for the catch.

Issue #270, opened 2024-02-14
waveOutPrepareHeader error in Blackthorne


Vort:
When playing Blackthorne game on Windows **host**, race condition in sound code may happen.
It results in one of three consequences:
1. `waveOutPrepareHeader(): error = 5` messages are displayed.
2. Sound output stops _(was triggered in the video)_.
3. Bochs crashes.

Problem reproduces most reliably in such scenario:
1. Orc shoots player.
2. Orc starts laughing.
3. Player jumps up the ledge.

https://github.com/bochs-emu/Bochs/assets/1242858/7781460e-de86-4f9f-b9a2-65f94f2cd7bb

During its execution, game switches between 11kHz and 22kHz output, which generates race condition between `bx_soundlow_waveout_win_c::set_pcm_params` (`waveOutOpen`) and `bx_soundlow_waveout_win_c::output` (`waveOutPrepareHeader`) functions.

Test files: [bthorne.zip](https://github.com/bochs-emu/Bochs/files/14277022/bthorne.zip).
Version: 1ff88fdd05a9b6640029fa0ba04304d58e833b3e.

vruppert:
Please try again with latest code.

Vort:
I still see problems # 1 and # 2 with 2979ba4. I wasn't able to get crash however.

Vort:
Race condition happens at exactly the same place as before.
Most likely, because of samples made by `fmopl_generator`.

Vort:
I decided to make small program demonstrating arbitrary resampling.
Here is source code [ArbResamp_src.zip](https://github.com/bochs-emu/Bochs/files/14864011/ArbResamp_src.zip), binary and data [ArbResamp_bin.zip](https://github.com/bochs-emu/Bochs/files/14864015/ArbResamp_bin.zip).
Program is made in C#, but I hope it is similar enough to C++ to be understood properly.
<details><summary>Source code</summary>

```cs
using System;
using System.Collections.Generic;
using System.IO;

namespace ArbResamp
{
    class Program
    {
        Program()
        {
            Resample("sine_1000", 2222, 3000);
            Resample("bt33", 11111, 44100);
        }

        void Resample(string fileIn, double sampleRate1, double sampleRate2)
        {
            Console.Write("Resampling " + fileIn + "...");

            double[] inData = SignalStoD(LoadWav(fileIn + ".wav"));

            double intCounter = 0.0;
            double intAdjust = sampleRate1 / sampleRate2;

[...]

Issue #107, opened 2023-10-22
When emm386 is loaded, log is flooded with "gate descriptor is not valid" messages


Vort:
Not sure if it is bug or not, but definitely a problem.
When I boot DOS 6.22 with `emm386.exe` specified in `config.sys`, Bochs starts lagging and flooding logs with messages like
```text
01104191449e[CPU0  ] interrupt(): gate descriptor is not valid sys seg (vector=0x1a)
01104191721e[CPU0  ] interrupt(): gate descriptor is not valid sys seg (vector=0x16)
01104191904e[CPU0  ] interrupt(): gate descriptor is not valid sys seg (vector=0x16)
01104192051e[CPU0  ] interrupt(): gate descriptor is not valid sys seg (vector=0x1a)
01104192241e[CPU0  ] interrupt(): gate descriptor is not valid sys seg (vector=0x16)
01104192433e[CPU0  ] interrupt(): gate descriptor is not valid sys seg (vector=0x28)
```
IPS drops from 75M to 0.09M (when logging to console, `-`) or 16M (when logging to `nul`).

Files needed to reproduce this problem are in archive [dos_emm.zip](https://github.com/bochs-emu/Bochs/files/13063367/dos_emm.zip).

Version: f455fa87e42eccbccf467e479183ed26605c8755.

stlintel:
try 

error: action=report, cpu=ignore

usually this error have meaning - this means the interrupt gate description is broken and usually it is developer problem.
I am wondering if DOS 6.22 was just broken because I am almost sure about this particular CPU code to be correct.

does it boot / work well if no logging messages ?

Vort:
> does it boot / work well if no logging messages ?

Looks like so.
with `error: action=report, cpu0=ignore` IPS is at ~50M, which is good.

![image](https://github.com/bochs-emu/Bochs/assets/1242858/4b219447-1fb9-4ecf-9112-208b34cc1844)

> I am wondering if DOS 6.22 was just broken

I don't know why it was made in a such way.
Maybe there was a reason, or maybe just mistake.

So what is the solution?
To recommend people disabling CPU logs before using `emm386.exe`?

stlintel:
Yes, I have no other idea. Welcome to any suggestion.
Today I tried to print  BX_ERROR before each fault like <#GP> here in order to provide enough information for people doing home-made OS development and debugging. This indicates that they made smth wrong. Till now I didn't have case when commercial software does smth wrong like this ...

[...]

PR #420, opened 2024-12-18
BIOS: Fix losing of interrupt flag (IF) after int13

humply:
Some programs expect the int 13h handler to return with interrupts enabled (IF).

Looking into a BIOS for ASUS A7A266 revealed int 13h enables interrupts and sets the IF on the stack.

Affected games:
- Blackthorne (MS-DOS 6.22) (will hang because of lost IF after int 13h).
- others...

### DIFF
--- a/bochs/bios/rombios.c
+++ b/bochs/bios/rombios.c
@@ -8905,6 +8905,7 @@ carry_set:
 ;   - make all called C function get the same parameters list
 ;
 int13_relocated:
+  sti                             ;; enable higher priority interrupts
 
 #if BX_ELTORITO_BOOT
   ;; check for an eltorito function
@@ -9049,6 +9050,14 @@ int13_out:
   pop ds
   pop es
   popa
+
+  ;; Note: Some DOS versions expect the int 13h handler to return with interrupts enabled (IF).
+  ;; Because of iret modify IF on stack.
+  push  bp
+  mov   bp, sp
+  or    byte ptr [bp+7], #0x02  ;; set interrupt flag (IF) on stack
+  pop   bp
+
   iret
 
 ;----------


fysnet:
I think this should be for hard drive services only. `DL >= 0x80`

A small addition to allow this would be to modify the line at 9011:
```
push #int13_out
  jmp _int13_diskette_function
```
to
```
push #int13_out_end
  jmp _int13_diskette_function
```
and then where the original `int13_out` was, add `int13_out_end`.

If you make this change, does the game still function as expected?

The notes I have say hard drive services only, not floppy services.

humply:
After trying out the suggested changes, so that the IF will not be touched for the `int13_diskette_function` case, the game would not start.

Unfortunately I couldn't find definitive information on what the proper behavior should be regarding int13 setting the IF, but the game does run on real hardware, so that's why I started looking into the BIOS of the hardware.

If it helps, the int13 entry point in the BIOS for the ASUS A7A266 starts with:
```
sti
pushf
cmp     dl, 1
ja      short loc_FAA79

popf
int     40h    ; Hard disk - Relocated Floppy Handler (original INT 13h)
retf    2

loc_FAA79:
popf
sti
push    bp
mov     bp, sp
sub     sp, 1Dh
push    bp
mov     bp, sp
add     bp, 2
push    ax
push    bx
push    cx
push    dx
push    ds
push    si
push    es
push    di
push    eax
or      word ptr [bp+23h], 200h   ; Set Interrupt Flag (IF) on stack
and     word ptr [bp+23h], 0FFFEh ; Clear Carry Flag (CF) on stack
mov     si, 0
mov     ds, si
cmp     ah, 0
jnz     short loc_FAAAB
call    sub_FAAD1
jb      short loc_FAAC0
jmp     short loc_FAAC4
...
```

Vort:
Small observation:
Game hangs with `himem.sys` 3.95 + `emm386.exe` 4.49, but not with `himemx.exe` 3.36 + `jemm386.exe` 5.79.

fysnet:
Okay, I guess leave it as is then. Thanks for checking.

vruppert (approved):

fysnet:
I added this fix to my BIOS as well.
As a favor, will you test the bios image at https://github.com/fysnet/i440fx (i440fx.bin) to see if it boots your game.
Or will you post a link to the image you boot the game so that I can test it.
Thank you.

Vort:
@fysnet here are game files: [Blackthorne (Black Hawk).English.zip](https://github.com/user-attachments/files/18204056/Blackthorne.Black.Hawk.English.zip). Add `himem.sys`, `emm386.exe` and bug should show up (with older versions I mean).

humply:
I tested the game with Bochs-2.8 and the `i440fx.bin` BIOS and the game started!

Because you're developing a BIOS, it's probably best to be able to reliably reproduce the issue yourself for further testing purposes.

fysnet:
> I tested the game with Bochs-2.8 and the `i440fx.bin` BIOS and the game started!

Thank you.

> here are game files

Thank you. I used the two older versions, the ones that come default with DOS 6.22, game works fine until a few minutes into it and Bochs simply quits. No `report to Microsoft` crash, nothing in the log indicating an error at the end of the log file, or anything like that. Bochs simply quits. It doesn't quit normal because the `.lock` file(s) are still there.

There were a lot, and I mean a lot, of the following line in the log file:
`interrupt(): gate descriptor is not valid sys seg (vector=0x16)`
The vector number varies.

Also, a lot, though not as many:
`interrupt(): soft_int && (gate.dpl < CPL)`

and then quite a few of:
`waveOutPrepareHeader(): error = 5`

I'm using the same `bochsrc` file as I did with VChess with an included `sb16` line.

If I find anything I will let you know.

Vort:
> Bochs simply quits.

Looks like bug.

> There were a lot, and I mean a lot, of the following line in the log file:

This is bug #107.
No one knows if it's a problem in emm386 or in Bochs.
I just switched to jemm386 because of this.

> and then quite a few of:

This is bug #270.
It is hard to fix because it requires implementation of [arbitrary resampling](https://github.com/bochs-emu/Bochs/issues/270#issuecomment-2036676669).

Issue #416, opened 2024-12-15
Virtua Chess fails to load from 800 MB partition


Vort:
When I try to load Virtua Chess from 800 MB partition with `BIOS-bochs-latest`, `16 bit code is too large` error appears. With `bios.bin-1.13.0`, game loads fine.

![bochs_vchess800](https://github.com/user-attachments/assets/348b95fe-0356-46c9-84de-618fefe660f5)

Test files: [VChess800.zip](https://github.com/user-attachments/files/18142240/VChess800.zip)
Version: be1a433

vruppert:
I can confirm this issue when using Bochs BIOS. With Ben's new BIOS the game also loads fine.

Vort:
Fixed with 908cea26ed18068594339805de8c5a3896408c20.

Issue #268, opened 2024-02-12
File error in Command & Conquer


Vort:
After Command & Conquer game is launched, it displays `FILE ERROR(CONQUER.INI): Bad file number`:
![bochs_cnc](https://github.com/bochs-emu/Bochs/assets/1242858/2cc974f7-c8e5-4079-af79-95265575393a)
It should complain about missing CD instead:
![vbox_cnc](https://github.com/bochs-emu/Bochs/assets/1242858/1b64b637-a184-4915-9bda-e973180f6be9)
Test files: [cnc.zip](https://github.com/bochs-emu/Bochs/files/14235327/cnc.zip).
Version: 30c1e74627f6664f3d3e2e259877c00ad20754d8.

vruppert:
This is a BIOS issue, since it works with SeaBIOS. I'll have a look at the INT 13h functions.

vruppert:
I haven't found the bug in Bochs BIOS yet, but I can confirm that it works correctly with Ben's BIOS (discussion #273).

Vort:
@vruppert I figured out why Command & Conquer glitches:
Problem occurs because `lcmpul` and `lsrul` functions damage `ebx` register.
There are more similar functions nearby, but such change is enough to make Command & Conquer start correctly:
```diff
diff --git a/bochs/bios/rombios.c b/bochs/bios/rombios.c
index d1ec39e1e..e35bfa18c 100644
--- a/bochs/bios/rombios.c
+++ b/bochs/bios/rombios.c
@@ -455,12 +455,14 @@ typedef unsigned long  Bit32u;
   ;; cmp function
   lcmpl:
   lcmpul:
+    push ebx
     and eax, #0x0000FFFF
     shl ebx, #16
     or  eax, ebx
     shr ebx, #16
     SEG SS
       cmp eax, dword ptr [di]
+    pop  ebx
     ret
 
   ;; sub function
@@ -519,6 +521,7 @@ typedef unsigned long  Bit32u;
 
   ;; sr function
   lsrul:
+    push ebx
     mov  cx,di
     jcxz lsr_exit
     and  eax, #0x0000FFFF
@@ -530,6 +533,7 @@ typedef unsigned long  Bit32u;
     mov  ebx, eax
     shr  ebx, #16
   lsr_exit:
+    pop  ebx
     ret
 
   ;; sl function 
```

vruppert:
Thanks for pointing this out. Fixed in Bochs repository now.

[...]

PR #421, opened 2024-12-18
BIOS: Fix register clobbering by int13_harddisk (#268)

humply:
Issue: The high words of EAX, EBX and EDX are clobbered by int13_harddisk.

This can be checked by e.g. `TEST/DOSTEST.EXE` included with HX DOS Extender v2.21.

Affected games:
- Blackthorne (MS-DOS 6.22) shows corrupted player animations.
- others...

See also commit 140e2e9, 49736dd and 4d5d311.

### DIFF
--- a/bochs/bios/rombios.c
+++ b/bochs/bios/rombios.c
@@ -478,13 +478,21 @@ typedef unsigned long  Bit32u;
   ;; mul function
   lmull:
   lmulul:
-    and eax, #0x0000FFFF
-    shl ebx, #16
-    or  eax, ebx
+    push edx
+    push ebx
+    push eax
+    and  eax, #0x0000FFFF
+    shl  ebx, #16
+    or   eax, ebx
     SEG SS
-    mul eax, dword ptr [di]
-    mov ebx, eax
-    shr ebx, #16
+    mul  eax, dword ptr [di]
+    mov  edx, eax
+    pop  eax
+    mov  ax, dx
+    shr  edx, #16
+    pop  ebx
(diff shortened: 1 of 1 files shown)


Vort:
I wonder how [I was able to run Blackthorne](https://github.com/bochs-emu/Bochs/issues/270) and not see these problems. Different version of the game maybe?

humply:
The version of Blackthorne used for testing was not the shareware version, but a full CD version with `BTHORNE.EXE` CRC32 `4BCED10D`.

Note that there are many other factors that can hide this issue as the OS and drivers can and will hook interrupt handlers.

You should be able to replicate the issue with the same version of the game running on MS-DOS 6.22 with only `HIMEM.SYS` and `EMM386.EXE` loaded (the game requires EMS).

Note: You will probably also need the interrupt flag fix to get the game to run (see my other pull request).

Vort:
Difference was not in game versions, but in partition size.
Looks like you fixed #416.

humply:
Great find! I can imagine that the helper functions that clobber the high words can have all kinds of effects.

Hopefully this fixes all the helper function clobbering issues (I did go over all of them).

Here is a screenshot of the corrupted player animations observed in Blackthorne without these fixes (the player animation is corrupted and can also disappear).
blackthorne

humply:
Maybe this also fixes the register clobbering observed with `call _int13_cdrom` (ebx is modified: BSD 5.2.1 boot loader problem)?

vruppert (approved):

PR #465, opened 2025-01-25
Xhci dbg fix

fysnet:
The function for some of the xHCI rings was never added to the USB_DEBUGGER code.
This should now allow you to see the DATA rings as well.
Use the `data=1` parameter:
```
  usb_debug: type=xhci, doorbell=1, event=1, data=1
```

### DIFF
--- a/bochs/config.cc
+++ b/bochs/config.cc
@@ -2,7 +2,7 @@
 // $Id$
 /////////////////////////////////////////////////////////////////////////
 //
-//  Copyright (C) 2002-2024  The Bochs Project
+//  Copyright (C) 2002-2025  The Bochs Project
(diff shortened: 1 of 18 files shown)


fysnet:
You can ignore the change in `win32res.rc`. For a reason I have yet to investigate, my Windows build won't build since it hasn't defined `BX_USB_DEBUGGER` yet, but at the top of the same file is:
```
#include "config.h"
```
So I will just need to find out why.

fysnet:
Volker, you will need to add the necessary changes to the GTK file(s) you have to get this PR to work with the GTK environment.

vruppert (approved):

PR #476, opened 2025-02-06
Vmware3

fysnet:
If the vmware3 class doesn't find a valid vmware3 image file, it will exit without removing the .lock file. A simple addition of
```
  bx_close_image(file_descriptor, pathname);
```
at line 257 would fix it:
https://github.com/bochs-emu/Bochs/blob/57020b856d88c2316ea0e77fb4a63d0b2b9c7e31/bochs/iodev/hdimage/vmware3.cc#L255-L259

However, to be more consistent with vmware4 I have tried to add a more consistent solution. I did not check to see if this works with a valid vmware3 file. I have none to test with. Please test...

### DIFF
--- a/bochs/iodev/hdimage/vmware3.cc
+++ b/bochs/iodev/hdimage/vmware3.cc
@@ -271,6 +271,7 @@ int vmware3_image_t::open(const char* _pathname, int flags)
   }
 
   bx_close_image(file_descriptor, pathname);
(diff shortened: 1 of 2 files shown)


vruppert:
I haven't found an image for testing, but it compiles and looks okay to me.

vruppert (approved):

Issue #515, opened 2025-04-10
Add gui option to go over an instruction - next


TracerDS:
I know `next` already exists as a "console command".
However having that as a keybind and as a separate option next to the `Step` button would be really great.
Instead of typing `n` on each instruction you would just hit the button or press a key. Just like step.

stlintel:
I have created initial attempt to implement this - see pull request #517
Can you test it please ?

PR #517, opened 2025-04-10
added StepOver button for enhanced debugger for issue #515

stlintel:


stlintel:
The branch is still incomplete

-> GTK supported never tested and might require update
-> Windows support works well but still there are issues. 

For example is I press 'Step Over' command immediately after I start Bochs gui debugger as first ever action -> Bochs freezes for few seconds and then crash. If I start with pressing 'step' few times and then continue to 'step over' seems like step over works perfectly. Initial debugging shows that Bochs crash even without calling the bx_dbg_step_over function so crash is gui debugger specific.

vruppert (commented):
I don't know what exactly should happen at "step over". The GTK version compiles and starts normally. When I press "Step over" right after starting Bochs, the emulation starts just like pressing "Continue", but the gui does not grey out the registers. The dialog is no longer responding, but terminating Bochs with the power button still works.
When I do something else before pressing "Step Over" (e.g. continue and break), the command acts similar to the step command.

stlintel:
The step over moves to next instruction through calls and repeatable instructions.
For example if you send call instruction, regular step will go inside function and continue executing it 
If you do step over - you will skip entire function body will sub-fubctions it potentially calls and resume immediately the return from function.

stlintel:
Voler, can you take a look if you see a problem with  peessing step over immediatelly after bochs starts on windows ?
It should look the same as if you write 'next' I'm command line - registrers gray out and bochs keep running forever (continue mode). Ctrl-c should stop it successfully.
For me bochs freezes and then tasks manager suggests to kill bochs non-responding task. I wonder what happens there and why it is limited to 1st instructon only. May just forbid pressing step over on 1st instruction...

Vort:
> May just forbid pressing step over on 1st instruction...

Such solutions are kinda fine in a very obscure cases.
This is not what happens here - hangs and crashes with fast 100% reproducibility are usually easy to track.

> Voler, can you take a look if you see a problem with peessing step over immediatelly after bochs starts on windows ?

I can.

> I wonder what happens there and why it is limited to 1st instructon only.

Hang happens because UI thread is locked by `bx_dbg_continue_command`.

Look how `case CMD_CONT:` is done: there is no locking there, `ActivateMenuItem` exits immediately.

stlintel:
I've got a fix now - reimplemented 'step over' command exactly as F8.
No crash anymore
Please test, especially in gtk and if okay - merge !

Issue #506, opened 2025-03-29
Won't boot ReactOS version 0.4.14 or version 0.4.15


fysnet:
Bochs won't boot version 0.4.14 with either chipset.

When booting the 0.4.14 livecd (first option, no debug), this is what you get:

![Image](https://github.com/user-attachments/assets/b846b7c8-ac14-479c-b9c0-0f0fd3b36fea)

When booting the 0.4.14 livecd second option `livecd (debug)`, it won't get past:

![Image](https://github.com/user-attachments/assets/a9517ab5-f4e2-44e2-9952-a65416819c99)

(Volker, we had a similar issue with my BIOS (https://github.com/fysnet/i440fx/issues/6) but now it is Bochs' BIOS.

Trying version 0.4.15 has a similar same effect. (livecd, first option, no debug)

![Image](https://github.com/user-attachments/assets/43bb7dfd-5090-4acf-99c9-7a377f9a2d6b)

When booting the 0.4.15 livecd second option `livecd (debug)`, it freezes at a `jmp short $-2` instruction.

When using my BIOS, booting version 0.4.14 with either chipset, works just fine. However, I get a very similar effect as using `BIOS-bochs-latest` and booting version 0.4.15. Rather than freezing at a `jmp short $-2` instruction, mine triple faults.

There must have been a change made with ReactOS for the latter version for Bochs to no longer show a Serial Port. However, I haven't found out why yet. I have made a few tests and found a couple inconsistencies, but they have no affect on the new version of ReactOS.

I don't know if this is a ReactOS issue or a Bochs issue. However, Volker, you previously got ReactOS to boot, yes?

fysnet:
Stop 0x7B means it can't find the boot device.

https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/bug-check-0x7b--inaccessible-boot-device?redirectedfrom=MSDN

Vort:
> When booting the 0.4.14 livecd (first option, no debug), this is what you get:

Can't reproduce:

![Image](https://github.com/user-attachments/assets/103d7060-0151-498e-8d13-5e4cd54e991d)

Probably something unusual is set in your config. Here is mine for comparison: [ReactOS_LiveCD.zip](https://github.com/user-attachments/files/19524398/ReactOS_LiveCD.zip).

> Trying version 0.4.15 has a similar same effect. (livecd, first option, no debug)

I confirm problem with 0.4.15.

[...]

PR #528, opened 2025-04-20
fix the timing and submit order of commands

fysnet:
This PR should fix issue #506. It now reads from the slave drive. However, I did not try writing, since a CDROM is not writable.


vruppert:
Writes to disk are broken with these changes. You should not change cmd_rwcon to cmd_ssbm in line 257.

fysnet:
Try it now. It looks like the only change needs to be the wait time on the timer. Changed from 1000us to 1us.

vruppert (approved):

PR #529, opened 2025-04-20
fix and cleanup Read TOC

fysnet:
This fixes an error in the Read TOC code and cleans up the return data.
Once it is thoroughly tested, it can be cleaned up more by returning a stream of bytes instead of initializing each byte as it does now.

### DIFF
--- a/bochs/iodev/hdimage/cdrom.cc
+++ b/bochs/iodev/hdimage/cdrom.cc
@@ -2,7 +2,7 @@
 // $Id$
 /////////////////////////////////////////////////////////////////////////
 //
-//  Copyright (C) 2002-2021  The Bochs Project
+//  Copyright (C) 2002-2025  The Bochs Project
 //
 //  This library is free software; you can redistribute it and/or
 //  modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


bochs/iodev/hdimage/cdrom.cc
```diff
@@ -105,130 +105,172 @@ void cdrom_base_c::eject_cdrom()
...
+      p[3] = 0xA1;    // POINT = 0xA1
+      p[4] = 0x00;    // Min
+      p[5] = 0x00;    // Sec
+      p[6] = 0x00;    // Frame
+      p[7] = 0x00;    // Zero or (Hour/PHour)
+      p[8] = 0x01;    // PMin
+      p[9] = 0x00;    // PSec
+      p[11] = 0x00;    // PFrame
+
+      p += 11;
```
stlintel, line 222: 11 and not 12 ?
by doing p+= 11 you going to overwrite p[11]=0x00 for PFrame that just written above

stlintel (changes requested):

bochs/iodev/hdimage/cdrom.cc
fysnet: Yes, you are right. Sorry about that.

vruppert (approved):

stlintel (approved):

PR #539, opened 2025-05-03
Try to make the fdc more like real hardware

fysnet:
- ~Create a "hack" to counter ReactOS clearing the reset bit and never restoring it to normal operation~
- Modify the source to return in milliseconds and then convert to microseconds for the timer
- Set DOR to normal operation after a reset via the Data Rate Select Register (0x3F4)
- Adjust the timing value to a smaller value (makes the floppy access faster)
- Real hardware sets the LOCK bit on by default
- Real hardware does not set the CHANGE DISK bit at reset
- Modified the BX_DEBUG for the DOR to print decimal instead of hex (value is bool, 0 or 1)
- Fix the `Illegal size for FIFO` error. Zero is a legal size on input, which translate to a size of 1.
- Added Sector numbers to the BX_DEBUG output of the `floppy_xfer` call.
- Fixed the `DUMP_REGISTERS` command to show the EOT value.

### DIFF
--- a/bochs/iodev/floppy.cc
+++ b/bochs/iodev/floppy.cc
@@ -754,6 +754,12 @@ void bx_floppy_ctrl_c::write(Bit32u address, Bit32u value, unsigned io_len)
         // transition from NORMAL to RESET
         BX_FD_THIS s.main_status_reg = FD_MS_BUSY;
         BX_FD_THIS s.pending_command = FD_RESET; // RESET pending
+        // The following line allows ReactOS to recognize the floppy disk.
+        // ReactOS writes zero to the DOR when stopping the motor within its boot process
+        //   https://github.com/reactos/reactos/blob/master/boot/freeldr/freeldr/arch/i386/pc/pcdisk.c
+        //     DiskStopFloppyMotor() (At time of this writing, it is Line 780)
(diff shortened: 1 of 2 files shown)


fysnet:
I tested this PR with ReactOS (no change), MS-DOS, FreeDOS, and WinXP. All functioned as expected.
Please test with a *nix variant to be sure.

vruppert (changes requested):
I don't like the hack for ReactOS and I don't want to test this PR before it's removed. The ReactOS people have to fix their code to make it work on real hardware. I'll test the remaining changes soon.

fysnet:
> I don't like the hack for ReactOS and I don't want to test this PR before it's removed. The ReactOS people have to fix their code to make it work on real hardware. I'll test the remaining changes soon.

Agreed. (Changes made)

On a side note, Github commented that `floppy.cc` had different line endings and will change them to Windows Style `crlf`. However, the `.gitattributes` file states that they should be `lf` only, though I think it might be in error or out of date. Have a look at https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings. It shows a more recent example on how to make sure all files have the desired line endings.

PR #550, opened 2025-05-14
Add int13/0x0D

fysnet:
Simply adding service Int13h/0x0D
See https://github.com/fysnet/i440fx/issues/10
DOS 7.11 boots with rombios.c without this addition, but you may want to add it anyway.
I don't think there is anything different, as far as emulation is concerned, between service 0x00 and 0x0D.


vruppert (approved):

fysnet:
Sorry. I didn't compile it to make sure it was good. I did go see if there was another matching `case:` line, but I guess I missed it.

PR #558, opened 2025-06-08
Use system font on win32 dialog boxes

Zeus2459:
The win32 dialog box font is tiny on my 1920x1080 Windows laptop. 

dialog_before

I have made the font less tiny by using the default system font instead.

dialog_after

Left fallback for that one person that wants to build it on Windows XP.
Reason for fallback: [NONCLIENTMETRICS#Remarks](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-nonclientmetricsa#remarks)

### DIFF
--- a/bochs/gui/win32paramdlg.cc
+++ b/bochs/gui/win32paramdlg.cc
@@ -348,23 +348,32 @@ int AskFilename(HWND hwnd, bx_param_filename_c *param, char *buffer)
 
 void InitDlgFont(void)
 {
-  LOGFONT LFont;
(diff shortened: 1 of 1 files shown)


Vort:
> Left fallback for that one person that wants to build it on Windows XP.

<details><summary>Builds for Windows 95 are possible as well with few extra changes</summary>

bochs_ewj_w95

</details> 

> I have made the font less tiny by using the default system font instead.

There are several minor glitches visible with Windows 7 tests:
Memory Options:
image
Clock and CMOS:
image

Issue #625, opened 2025-10-02
Ubuntu 10.04 (not 10.04.4) does not show any text


Randrianasulu:
Git commit:

48eb46b329c627dd319f50c033588bd888537227

configure line:
```
/configure --enable-sb16 --enable-es1370 --enable-e1000 --enable-ne2000 --enable-usb --enable-clgd54xx --enable-voodoo --enable-all-optimizations --enable-geforce
````

Note that host is 32bit Slackware 15.0 (mostly) with 64-bit kernel

bochsrc:

[bochsrc.txt](https://github.com/user-attachments/files/22665577/bochsrc.txt)

bochsout.txt

[bochsout.txt](https://github.com/user-attachments/files/22665580/bochsout.txt)

Screenshot:

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/c4219a8e-3c09-4a5f-a2a5-32d64b850033" />

Vort:
This problem appears with 10.04.4 as well. I was making tests with `model=geforcefx5900` and was hoping that `model=geforce6800` is similar enough to not cause additional problems. I will try to find what went wrong.

Vort:
I doubt this change is entirely correct, but it should be enough for a while: #628.

Randrianasulu:
Fixed, thanks!

PR #628, opened 2025-10-04
Fix missing text in Ubuntu 10 with NV40

Vort:
This change fixes texture coordinates mapping for NV40 (issue #625).


Randrianasulu:
Yeah, it works here (manually applied this commit with git apply), thanks in advance!

vruppert (approved):

PR #637, opened 2025-10-11
Start of Audio CD-ROM access

fysnet:
*** Do not merge this PR into the master at this time. It is a work in progress and needs testing ***

See https://github.com/bochs-emu/Bochs/discussions/631 for more information.

### DIFF
--- a/bochs/iodev/hdimage/cdrom_win32.cc
+++ b/bochs/iodev/hdimage/cdrom_win32.cc
@@ -64,6 +64,34 @@ extern "C" {
 
 #include <stdio.h>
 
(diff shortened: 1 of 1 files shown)


vruppert:
I have some fixes for builds without MSVC here, but they are not complete:
For non-WIN32 builds:
Move endif in cdrom_win32.h to the last line in file.
For cross compilation for Windows on Linux:
File names are case sensitive, so first letter of include file name must be lower case.
Cross build still fails, since there are illegal cases: "(DWORD)(LPVOID)".

vruppert (approved):

Issue #678, opened 2025-11-26
Incorrect handling of 0x46 CDB commands (ATAPI)


disean:
Version: 08c8ed933916d90371a0d609f9321bc0d1679801
ISO File: https://github.com/reactos/reactos/actions/runs/19678484187/artifacts/4676568686 (livecd.iso)

When the host sends a 0x46 CDB command, which has allocation length of 8, bochs incorrectly returns more than 8 bytes of data, resulting in a host buffer overrun.

```diff
Send packet command
    IO 000001F6 <-- a0
    IO 000001F7 --> 50
    IO 000001F4 <-- 08
    IO 000001F5 <-- 00
    IO 000001F1 <-- 00
    IO 000001F7 <-- a0
    
Wait for ready for CDB
    IO 000001F7 --> 58
    IO 000001F2 --> 01 (IR = 09, phase = accept CDB)

Send CDB 46 (data read)
    IO 000001F0 <-- 00000046 Cdb[0] I/O 32
    IO 000001F0 <-- 00000000 Cdb[1] I/O 32
    IO 000001F0 <-- 00000008 Cdb[2] I/O 32 (allocation length = 8)
    
Interrupt
    IO 0000C002 --> 04
    IO 0000C002 <-- 04 (clear BM DMA)
    
    IO 000001F7 --> 58
    IO 000001F2 --> 02 (IR = 0A, phase = data read)
    
    IO 000001F4 --> 08
    IO 000001F5 --> 00 (data size = 8 bytes)

Read data (8 bytes)
    IO 000001F0 --> 48000000 Cdb[0] I/O 32
    IO 000001F0 --> 08000000 Cdb[1] I/O 32

Interrupt
    IO 0000C002 --> 04
    IO 0000C002 <-- 04 (clear BM DMA)
    
    Incorrect phase transition goes here:
-   IO 000001F7 --> 58
-   IO 000001F2 --> 02 (IR = 0A, phase = data read)

-   IO 000001F4 --> 08
-   IO 000001F5 --> 00 (data size = 8 bytes)
    
    Expected behavior:
+   IO 000001F7 --> 50
+   IO 000001F2 --> 03 (IR = 03, phase = status)
```

fysnet:
The following comment (from me) probably proves your idea:
https://github.com/bochs-emu/Bochs/blob/08c8ed933916d90371a0d609f9321bc0d1679801/bochs/iodev/harddrv.cc#L2079-L2081

[...]

PR #683, opened 2025-12-01
Atapi 46 fix

fysnet:
@disean please check this to see if it fixes your issue at #678
None of the test suites I have (WinXP, etc.) call this command, so I don't have any test cases, though I think this fix is correct.


vruppert:
I just had a look at this PR and I need to know which live CD has been used for testing. The live CD of release 0.4.15 boots in normat mode with both Bochs master and this PR applied. The latest liive CD (nightly build) fails with both version. I just need to know that live CD revision X in mode Y works with this PR applied, but not with master. If everything is okay, then I'd like to include it in Bochs release 3.1.

Vort:
I will explain it in a few minutes.

Vort:
> I just need to know that live CD revision X in mode Y works with this PR applied, but not with master.

To access Live CD mode in new versions of ReactOS, **bootcd** iso needs to be downloaded, but then, in the menu, "ReactOS Live Environment (Graphics Mode)" should be selected.

Here is the version which fails with blue screen without this PR and works fine with it:
[reactos-bootcd-0.4.17-dev-682-gc7e1c79-x86-gcc-lin-dbg.iso](https://iso.reactos.org/bootcd/reactos-bootcd-0.4.17-dev-682-gc7e1c79-x86-gcc-lin-dbg.7z)

ros_livecd

vruppert (approved):

PR #716, opened 2026-01-13
Round adaptive block size to the next highest power of 2

Vort:
This change fixes `[MEM0  ] Block size 4352 is not power of two !`  crash when using `memory: guest=17, host=17` configuration.

### DIFF
--- a/bochs/main.cc
+++ b/bochs/main.cc
@@ -105,6 +105,17 @@ BOCHSAPI BX_MEM_C bx_mem;
 
 char *bochsrc_filename = NULL;
 
+Bit32u round_up_to_power_of_2(Bit32u x)
+{
+  Bit32u r = x - 1;
+  r |= r >> 1;
+  r |= r >> 2;
+  r |= r >> 4;
+  r |= r >> 8;
+  r |= r >> 16;
+  return r + 1;
+}
+
 size_t bx_get_timestamp(char *buffer)
 {
 #if VER_DEVFLAG == 1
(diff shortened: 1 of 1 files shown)


fysnet:
Though it doesn't really make much difference, being this algorithm is nice and quick, it doesn't work for a value less than 2.
Maybe a comment should be added to indicate that this routine assumes a value of 2 or greater.
Maybe a check for a value less than 2 could be added, but then that defeats the purpose of being quick.

stlintel:
> Though it doesn't really make much difference, being this algorithm is nice and quick, it doesn't work for a value less than 2. Maybe a comment should be added to indicate that this routine assumes a value of 2 or greater. Maybe a check for a value less than 2 could be added, but then that defeats the purpose of being quick.

Value less than 2 is what ?
1 or 0 ?
1 is also power if 2 for me
And 0 is special, actually I would prefer to round 0 to 0

Vort:
> Though it doesn't really make much difference

Yes, because of this line right before `round_up_to_power_of_2` call:
`if (memBlockSize < 4096) memBlockSize = 4096;`

> it doesn't work for a value less than 2

It rounds 0 to 0.

> Maybe a comment should be added

Maybe. However, if someone decides to use this function somewhere else, he will make tests and will notice if 0 -> 0 case breaks something.

> but then that defeats the purpose of being quick.

I think whether such check is needed should be decided in cases where both rounding of 0 and performance are important. It's a future problem.

PR #731, opened 2026-02-17
Implement environment bump mapping for GeForce

Vort:
This change allows [BumpWaves](https://github.com/user-attachments/files/25371854/BumpWaves.zip) test to work correctly with 45.23 driver, NV20 and NV35.
Additionally, this change fixes rendering of hardware accelerated cube from DxDiag 4.08 with 45.23 driver and NV35.

Screenshot_2026-02-17_20-02-06

### DIFF
--- a/bochs/iodev/display/geforce.cc
+++ b/bochs/iodev/display/geforce.cc
@@ -4067,22 +4067,8 @@ float rc_get_var(Bit32u cw, Bit32u shift, float regs[16][4], Bit32u civ)
   }
 }
 
-void bx_geforce_c::d3d_register_combiners(gf_channel* ch, float ps_in[16][4], float out[4])
+void bx_geforce_c::d3d_register_combiners(gf_channel* ch, float regs[16][4], float out[4])
 {
-  float regs[16][4];
-  for (Bit32u ci = 0; ci < 4; ci++) {
-    regs[0][ci] = 0.0f;
-    regs[3][ci] = 1.0f;
-    regs[4][ci] = ps_in[1][ci];
-    regs[5][ci] = ps_in[2][ci];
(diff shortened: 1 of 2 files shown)


stlintel (approved):

sskras:
@Vort, welcome back – yay!

That looks like a complex piece of code. I wonder how do you reverse/manage to get all these 3D graphics details right... Looks mind-blowing to me. My warmest cheers :)

Vort:
@sskras, thanks.
While figuring out why DxDiag 4.08 produces wrong colors with NV35, I noticed that register combiners were using uninitialized data from texture register. It happened because DxDiag 4.08 cube is not textured; however, register needed to be filled somehow anyway. For a long time I noticed cases, where both register combiners and pixel shaders were active, but didn't knew how to combine their results. Now I tried to fill texture registers with data from pixel shaders and it gave correctly colored cube.
Having pixel shaders filling texture registers made me wonder about how these registers should be filled with NV20, which has no pixel shaders. For some effects, like environment bump mapping, it is not possible to just sample single texture. Because xemu implements similar GPU, I looked for how texture registers are filled in its source code. It was possible to just implement something similar, but I wanted to have more information, so I also checked Mesa source code and found precise name of mechanism, used in such effects. Quick search revealed quite detailed document, [NV_texture_shader](https://registry.khronos.org/OpenGL/extensions/NV/NV_texture_shader.txt), which confirmed formulas found in xemu code.
Unfortunately, sometimes detailed information is not available and results have to be obtained by experimentation, disassembling and guessing. For example, details about how near plane clipping works are not clear. I made several attempts at implementing it right, but each time examples appeared which showed that my guesses were wrong.

PR #742, opened 2026-03-05
Fix critical 64-bit address truncation, PCI hole handling, and icache bugs

Kharosx0:
## Summary

This PR fixes bugs  when running guests with >4GB RAM.  Multiple code paths still silently truncate 64-bit physical addresses to 32 bits or fail to account for the PCI MMIO hole (0xC0000000–0xFFFFFFFF) that creates a non-contiguous physical-to-linear address mapping.

### 1. Instruction cache (icache) write-stamp table truncates 64-bit physical addresses

The `bxPageWriteStampTable` class used `((Bit32u) pAddr) >> 12` to hash physical addresses into page slots. For physical addresses above 4GB, this truncates the upper 32 bits, causing different physical pages to alias into the same write-stamp slot. This leads to missed self-modifying code (SMC) detection and stale icache entries — potentially executing old instructions after code has been modified.

The `markICache()` and `decWriteStamp(unsigned len)` methods also used `PAGE_OFFSET((Bit32u) pAddr)` to compute sub-page offsets, which has the same truncation problem — the wrong 128-byte region gets marked/invalidated.

**Fix:** `hash()` now translates through `bx_translate_gpa_to_linear()` before shifting, a new bounds-checked `index_of()` method replaces raw `hash()` calls for array indexing, and all page-offset calculations use 64-bit masking. The table is expanded from 1M entries (4GB) to 2M entries (8GB linear space) to cover the full post-PCI-hole address range. Added `len == 0` early-return guards to prevent undefined shifts.

**Files:** `cpu/icache.h`

---

### 2. `serveICacheMiss` truncates physical address page offset

`PAGE_OFFSET((Bit32u) pAddr)` in `serveICacheMiss()` wraps addresses above 4GB, producing wrong page offsets for instruction fetch.

**Fix:** Replaced with `(Bit32u)(pAddr & (bx_phy_address)0xfff)`.

**Files:** `cpu/icache.cc`

---

### 3. `writePhysicalPage` passes wrong address to `decWriteStamp`

Both `memory_stub.cc` and `memory.cc` implementations of `writePhysicalPage()` passed the post-PCI-hole-translated `linear_addr` to `decWriteStamp()`. Since `decWriteStamp()` now translates internally via `hash()`, this caused double translation, indexing into the wrong write-stamp slot and silently breaking SMC detection for memory above the PCI hole.

**Fix:** Introduced a separate `phys_addr` cursor (initialized from `a20addr`) for all `decWriteStamp()` calls, while continuing to use `linear_addr` for `get_vector()` RAM access. Both cursors are incremented in the chunked-write loop.

**Files:** `memory/memory_stub.cc`

---

### 4. MWAIT `check_monitor` / `is_monitor` uses wrong address

`check_monitor()` and `is_monitor()` were called with the post-PCI-hole `linear_addr` instead of the physical `a20addr`. Since MWAIT monitors physical pages, this caused the monitor to trigger on the wrong page for addresses above the PCI hole, or miss triggers entirely.

**Fix:** All three call sites (`memory.cc`, `memory_stub.cc` `writePhysicalPage`, `memory_stub.cc` `getHostMemAddr`, `misc_mem.cc` `getHostMemAddr`) now use `a20addr`.

**Files:** `memory/memory.cc`, `memory/memory_stub.cc`, `memory/misc_mem.cc`

---

### 5. `dbg_fetch_mem` / `dbg_set_mem` don't handle PCI hole

The debugger memory access functions used raw `a20addr` directly with `get_vector()`, which indexes into the RAM backing store. For addresses in or above the PCI hole, this accesses the wrong memory or goes out of bounds. They also used bare `a20addr++` which skips the A20 gate on wrap.

**Fix:** Added `bx_is_pci_hole_addr()` checks, `bx_translate_gpa_to_linear()` for RAM access, and `A20ADDR()` wrapping on address increment. `dbg_set_mem` uses a two-pass approach (validate all, then write) to prevent partial writes on error.

**Files:** `memory/memory_stub.cc`

---

### 6. `dbg_crc32` produces wrong results and truncates addresses

Three bugs: (a) used `unsigned` for address arithmetic, truncating ranges above 4GB; (b) called `crc32()` per chunk which resets the CRC seed each iteration, producing incorrect multi-page CRC results; (c) didn't account for the PCI hole.

**Fix:** Complete rewrite using `bx_phy_address` (64-bit) for all arithmetic, new `crc32_extend()` function for seeded CRC accumulation, PCI hole validation, and `bx_translate_gpa_to_linear()` for RAM access.

**Files:** `memory/memory_stub.cc`, `crc.cc`, `bochs.h`

---

### 7. `prefetch` sanity check doesn't handle PCI hole addresses

The prefetch error path compared raw physical addresses against `get_memory_len()` without accounting for the PCI hole gap, producing misleading panic messages or incorrect "bogus memory" errors for legitimate above-4GB addresses.

**Fix:** Added explicit `bx_is_pci_hole_addr()` check and `bx_translate_gpa_to_linear()` before the length comparison. Error messages now include both physical and linear addresses for easier diagnosis.

**Files:** `cpu/cpu.cc`

---

### 8. `get_num_blocks` silently truncates on large memory configs

`Bit64u len / Bit32u block_size` was assigned to a `Bit32u`, silently truncating the result for memory configurations where the block count exceeds 2^32.

**Fix:** Use `Bit64u` intermediate with `BX_ASSERT` guard.

**Files:** `memory/memory_stub.cc`

---

### 9. `init_memory` block pointer arithmetic overflows at 4GB

`idx * block_size` uses `unsigned * Bit32u` which overflows to zero at block index 256 with 16MB blocks (256 × 16MB = 4GB).

**Fix:** Cast to `Bit64u(idx)` before multiplication.

**Files:** `memory/memory_stub.cc`

---

### 10. `load_RAM` uses `unsigned long` and doesn't handle PCI hole

`load_RAM()` used `unsigned long` for size and offset (32-bit on many platforms), and loaded directly via `get_vector(offset)` without PCI hole translation, corrupting memory for images loaded above the PCI hole boundary. The caller in `main.cc` also used `get()` (returns `Bit32s`) instead of `get64()`, truncating addresses above 2GB.

**Fix:** Use `Bit64u` for size, `bx_phy_address` for GPA, page-chunked loading with PCI hole checks and `bx_translate_gpa_to_linear()`. Changed `main.cc` call to `get64()`. Proper `close(fd)` on all error paths.

**Files:** `memory/misc_mem.cc`, `main.cc`

---

### 11. Minor: APIC format string mismatch

`len` parameter is `unsigned` but format string used `%d` (signed).

**Fix:** Changed to `%u`.

**Files:** `cpu/apic.cc`

---

### 12. Minor: `bx_asize_mask` narrowing warning

`BX_CONST64(0xffffffffffffffff)` assigned to `bx_address` array without explicit cast produces MSVC narrowing warnings when `bx_address` is 64-bit.

**Fix:** Added explicit `(bx_address)` cast.

**Files:** `cpu/access.cc`

---

### `crc32_extend(Bit32u seed_crc, const Bit8u *buf, int len)`

Incremental CRC-32 that continues from a previously finalized CRC value. Required for computing CRC across non-contiguous memory regions (e.g., page-chunked address ranges that skip the PCI hole). The existing `crc32()` function is now a thin wrapper calling `crc32_extend(0, ...)`.

**Files:** `crc.cc`, `bochs.h`

### DIFF
--- a/bochs/memory/memory_stub.cc
+++ b/bochs/memory/memory_stub.cc
@@ -77,16 +77,12 @@ Bit64u BX_MEMORY_STUB_C::get_memory_len(void)
   return (BX_MEM_THIS len);
 }
 
-// Use Bit64u intermediate to prevent silent truncation when len > 4GB,
-// since Bit64u / Bit32u would otherwise implicitly truncate to Bit32u.
 Bit32u BX_MEMORY_STUB_C::get_num_blocks()
 {
-  Bit64u num_blocks64 = BX_MEM_THIS len / BX_MEM_THIS block_size;
+  Bit32u num_blocks = BX_MEM_THIS len / BX_MEM_THIS block_size;
   if ((BX_MEM_THIS len % BX_MEM_THIS block_size) != 0)
-    num_blocks64++;
-
-  BX_ASSERT(num_blocks64 <= 0xffffffffu);
-  return (Bit32u) num_blocks64;
+    num_blocks++;
+  return num_blocks;
 }
 
 Bit8u* BX_MEMORY_STUB_C::alloc_vector_aligned(Bit64u bytes, Bit64u alignment)


Vort:
No need to touch `get_num_blocks`: 1. Assert is already done in [init_memory](https://github.com/bochs-emu/Bochs/blob/86e5508783526802bedc8c288e358bf20f673973/bochs/memory/memory_stub.cc#L137). 2. `block_size` can't be lower than 4K, which means overflowing is practically impossible.

Kharosx0:
> No need to touch `get_num_blocks`: 1. Assert is already done in [init_memory](https://github.com/bochs-emu/Bochs/blob/86e5508783526802bedc8c288e358bf20f673973/bochs/memory/memory_stub.cc#L137). 2. `block_size` can't be lower than 4K, which means overflowing is practically impossible.

Ah true, reverted that change.

stlintel:
> 2. serveICacheMiss truncates physical address page offset
> PAGE_OFFSET((Bit32u) pAddr) in serveICacheMiss() wraps addresses above 4GB, producing wrong page offsets for instruction fetch.
> 
> Fix: Replaced with (Bit32u)(pAddr & (bx_phy_address)0xfff).
> 
> Files: cpu/icache.cc
> 

PAGE_OFFSET(addr) was exactly the replacement of using magic constant 0xfff.
What is the difference here ? Who cares if address wraps above 4GB if we care only for low 12 bit of it ?
Even with bx_translate_gpa_to_linear there is no change is lower 12 bits of the address.

Similarly Bit32u() truncation is applied here for speed in 32-bit hosts. It doesn't affect lower 12-bit but forces entire operation to 32-bit instead of 64-bit.

stlintel:
About PageWriteStampTable: you wrote

> 1. Instruction cache (icache) write-stamp table truncates 64-bit physical addresses
> The bxPageWriteStampTable class used ((Bit32u) pAddr) >> 12 to hash physical addresses into page slots. For physical addresses above 4GB, this truncates the upper 32 bits, causing different physical pages to alias into the same write-stamp slot. This leads to missed self-modifying code (SMC) detection and stale icache entries — potentially executing old instructions after code has been modified.
> 
> The markICache() and decWriteStamp(unsigned len) methods also used PAGE_OFFSET((Bit32u) pAddr) to compute sub-page offsets, which has the same truncation problem — the wrong 128-byte region gets marked/invalidated.
> 
> Fix: hash() now translates through bx_translate_gpa_to_linear() before shifting, a new bounds-checked index_of() method replaces raw hash() calls for array indexing, and all page-offset calculations use 64-bit masking. The table is expanded from 1M entries (4GB) to 2M entries (8GB linear space) to cover the full post-PCI-hole address range. Added len == 0 early-return guards to prevent undefined shifts.
> 
> Files: cpu/icache.h

The PageWriteStampTable is okay to be truncated. It wont' cause missing SMC, it might potentially result in false SMC detection. fineGranularityMapping is acting within 4KB boundary (and therefore absolutely not afffected with all >4GB games). This is a bit mask, a bit is set there each time a code fetch allocates 'a 128-byte cache line' into code cache.
Every incoming store is checked against fineGranularityMapping and if hit - may result in SMC flush.

Now PageWriteStampTable itself is just a hash. All that I need is that all loads and all stores and all memory operation with the same physical address will be mapped to the same entry. Regardless how they are organized. So CPU doesn't care for hole address in access.cc while calling decWriteStamp, so you in memory should do the same - and all will be just fine.
This means decWriteStamp calls must be done in A20ADDR and not in 'linear_addr'

If this PHY_MEM_PAGES_IN_4G_SPACE or PHY_MEM_PAGES_IN_8G_SPACE less important. Might be increased to PHY_MEM_PAGES_IN_8G_SPACE if you want, this would be enough to prevent overflow.

  BX_CPP_INLINE static Bit32u hash(bx_phy_address pAddr) {
    // can share writeStamps between multiple pages if >32 bit phy address  // see, it already fine to share writeStamps 
    return ((Bit32u) pAddr) >> 12;
  }

stlintel:
Meanwhile I have pushed most critical fixes from the PR: 27bd71abef5846b7f8f9a34db296005a1d345545 and 5857f9462f67a33a7ce0ff72a6236e929108fc83

stlintel:
As you may notice - you fall into the same trap of putting to much stuff into single PR.
And it cause you to rewrite it completely multiple times based on comments, instead of merging relevant parts quickly.
I see at least 4, may be more different PR here ...

I would suggest separate PR for 
5. dbg_fetch_mem / dbg_set_mem don't handle PCI hole
6. dbg_crc32 produces wrong results and truncates addresses
10. load_RAM uses unsigned long and doesn't handle PCI hole

3 and 4 are merged
1 and 2 most likely wrong and should be discarded

Kharosx0:
Closed and split to https://github.com/bochs-emu/Bochs/pull/747 https://github.com/bochs-emu/Bochs/pull/746 https://github.com/bochs-emu/Bochs/pull/745

PR #756, opened 2026-04-02
Remove 64-bit from Base Address

fysnet:
This PR adds two small differences.
1) Removes the 64-bit Base Address flag and defaults to 32-bit using the `ADDR_CAP_64` define.
2) Using the `PORT_POWER_CTRL` define, can now use power-all-the-time, or power-by-port-with-software.


fysnet:
Your fixes and additions look good. Thank you.

PR #759, opened 2026-04-08
Fix VMX on UEFI

Kharosx0:
This fixes Intel VMX initialization when booting through OVMF.

Bochs BIOS enables VMX in IA32_FEATURE_CONTROL itself, but the UEFI/fw_cfg path does not. As a result, guests that rely on VMX can fail under UEFI even though the same setup works with the Bochs BIOS.

This change adds a small CPU helper that ensures the required IA32_FEATURE_CONTROL bits are present and calls it from the fw_cfg reset path before external firmware runs.

What this changes:
- sets the required bits when VMX is enabled
- limits the workaround to the external firmware path (UEFI), where the missing initialization occurs

With this change, VMX is properly recognized and used e.g. by Hyper-V. Without the change, VMX is unused.

### DIFF
--- a/bochs/cpu/cpu.h
+++ b/bochs/cpu/cpu.h
@@ -4852,6 +4852,7 @@ public: // for now...
 #endif
 
   BX_SMF void reset(unsigned source);
+  BX_SMF void allowVmxForFirmware(void);
   BX_SMF void shutdown(void);
   BX_SMF void enter_sleep_state(unsigned state);
   BX_SMF void handleCpuModeChange(void);
--- a/bochs/cpu/cpu.h
+++ b/bochs/cpu/cpu.h
@@ -4852,7 +4852,6 @@ public: // for now...
 #endif
 
   BX_SMF void reset(unsigned source);
-  BX_SMF void allowVmxForFirmware(void);
   BX_SMF void shutdown(void);
   BX_SMF void enter_sleep_state(unsigned state);
   BX_SMF void handleCpuModeChange(void);
@@ -5395,6 +5394,7 @@ public: // for now...
   BX_SMF void VMexit_CR8_Read(bxInstruction_c *i) BX_CPP_AttrRegparmN(1);
   BX_SMF void VMexit_CR8_Write(bxInstruction_c *i) BX_CPP_AttrRegparmN(1);
   BX_SMF void VMexit_DR_Access(unsigned read, unsigned dr, unsigned reg);
+  BX_SMF void allowVmxForFirmware(void);
 #if BX_SUPPORT_VMX >= 2
   BX_SMF void Virtualization_Exception(Bit64u qualification, Bit64u guest_physical, Bit64u guest_linear);
   BX_SMF void vmfunc_eptp_switching(void);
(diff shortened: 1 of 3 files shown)


bochs/cpu/init.cc
```diff
@@ -1307,6 +1307,24 @@ void BX_CPU_C::reset(unsigned source)
   BX_INSTR_RESET(BX_CPU_ID, source);
 }
 
+void BX_CPU_C::allowVmxForFirmware(void)
+{
+#if BX_SUPPORT_VMX
```
stlintel: please make this function completely under BX_SUPPORT_VMX
then put if BX_SUPPORT_VMX around place you call it.

bochs/cpu/init.cc
```diff
@@ -1307,6 +1307,24 @@ void BX_CPU_C::reset(unsigned source)
...
+  // The Bochs BIOS enables VMX in IA32_FEATURE_CONTROL itself. External
+  // firmware uses the fw_cfg path and needs the same setup from the emulator.
+  if (! is_cpu_extension_supported(BX_ISA_VMX)) {
+    return;
+  }
+
+  const Bit32u required_bits = BX_IA32_FEATURE_CONTROL_BITS;
```
stlintel: this check is redundant. The OR is enough, it will set the bits regardless if they were set or clear.
just use the OR only

bochs/iodev/fw_cfg.cc
```diff
@@ -232,6 +233,15 @@ void bx_fw_cfg_c::reset(unsigned type)
   cur_entry = FW_CFG_INVALID;
   cur_offset = 0;
   dma_addr = 0;
+
+  // The Bochs BIOS enables VMX on its own. External firmware uses fw_cfg,
+  // whose reset callback runs after CPU reset, so allow VMX here first.
+  if (SIM->get_param_bool(BXPN_FW_CFG_ENABLED)->get()) {
```
stlintel: just place entirely all this under BX_SUPPORT_VMX
so if no VMX compiled in - no attempt to read param, also no for over BX_SMP_PROCESSORS

side Q, why reading of param is needed ?
why this function is ever called if fw_cfg is not enabled ?

stlintel (changes requested):

Kharosx0:
Addressed comments

bochs/cpu/init.cc
Kharosx0: Addressed your comments, ty

bochs/cpu/cpu.h
```diff
@@ -5394,6 +5394,7 @@ class BX_CPU_C : public logfunctions {
   BX_SMF void VMexit_CR8_Read(bxInstruction_c *i) BX_CPP_AttrRegparmN(1);
   BX_SMF void VMexit_CR8_Write(bxInstruction_c *i) BX_CPP_AttrRegparmN(1);
   BX_SMF void VMexit_DR_Access(unsigned read, unsigned dr, unsigned reg);
+  BX_SMF void allowVmxForFirmware(void);
```
stlintel, line 5397: Ifdef also here

stlintel (changes requested):

bochs/cpu/cpu.h
Kharosx0: Isn't it already under the appropriate ifdef here? Hence why we also see VMexit* APIs defined?

stlintel: Ah sorry, missed it

stlintel (approved):

PR #770, opened 2026-06-13
uhci endianness

fysnet:
Attempt to fix UHCI' host endianness.
This is a test, not knowing if this is the fix or not.

### DIFF
--- a/bochs/iodev/usb/uhci_core.cc
+++ b/bochs/iodev/usb/uhci_core.cc
@@ -726,7 +726,12 @@ void bx_uhci_core_c::uhci_timer(void)
         }
 
         // read in the queue
+#ifdef BX_LITTLE_ENDIAN
         DEV_MEM_READ_PHYSICAL(item & ~0xF, sizeof(struct QUEUE), (Bit8u *) &queue);
+#else
+        DEV_MEM_READ_PHYSICAL((item & ~0xF) + 0, sizeof(Bit32u), &queue.horz);
+        DEV_MEM_READ_PHYSICAL((item & ~0xF) + 4, sizeof(Bit32u), &queue.vert);
+#endif
 
         // this massively populates the log file, so I keep it commented out
         //BX_DEBUG(("Queue at 0x%08X:  horz = 0x%08X, vert = 0x%08X", item & ~0xF, queue.horz, queue.vert));
@@ -752,7 +757,14 @@ void bx_uhci_core_c::uhci_timer(void)
 
       // else, we found a Transfer Descriptor
       address = item & ~0xF;
+#ifdef BX_LITTLE_ENDIAN
       DEV_MEM_READ_PHYSICAL(address, sizeof(struct TD), (Bit8u *) &td);
+#else
+      DEV_MEM_READ_PHYSICAL(address +  0, sizeof(bit32u), &td.dword0);
+      DEV_MEM_READ_PHYSICAL(address +  4, sizeof(bit32u), &td.dword1);
+      DEV_MEM_READ_PHYSICAL(address +  8, sizeof(bit32u), &td.dword2);
+      DEV_MEM_READ_PHYSICAL(address + 12, sizeof(bit32u), &td.dword3);
+#endif
       const bool depthbreadth = (td.dword0 & 0x0004) ? 1 : 0;     // 1 = depth first, 0 = breadth first
       const bool is_active = (td.dword1 & (1<<23)) > 0;
       bool was_short = 0, was_stall = 0;
@@ -989,6 +1001,7 @@ bool bx_uhci_core_c::DoTransfer(Bit32u address, struct TD *td)
       case USB_TOKEN_OUT:
       case USB_TOKEN_SETUP:
         if (maxlen > 0) {
+          // no need to endian this. It is little endian no matter what
           DEV_MEM_READ_PHYSICAL_DMA(td->dword3, maxlen, p->packet.data);
         }
         ret = broadcast_packet(&p->packet);
@@ -1016,6 +1029,7 @@ bool bx_uhci_core_c::DoTransfer(Bit32u address, struct TD *td)
         ret = USB_RET_BABBLE;
       }
       if (len > 0) {
+        // no need to endian this. It is little endian no matter what
         DEV_MEM_WRITE_PHYSICAL_DMA(td->dword3, len, p->packet.data);
       }
     } else {
--- a/bochs/iodev/usb/uhci_core.cc
+++ b/bochs/iodev/usb/uhci_core.cc
@@ -2,7 +2,7 @@
 // $Id$
 /////////////////////////////////////////////////////////////////////////
 //
-//  Copyright (C) 2009-2025  Benjamin D Lunt (fys [at] fysnet [dot] net)
+//  Copyright (C) 2009-2026  Benjamin D Lunt (fys [at] fysnet [dot] net)
 //                2009-2026  The Bochs Project
 //
 //  This library is free software; you can redistribute it and/or
@@ -363,11 +363,6 @@ Bit32u bx_uhci_core_c::read(Bit32u address, unsigned io_len)
   // don't flood the log with reads from the Frame Register
   if (offset != 0x06)
     BX_DEBUG(("register read from address 0x%04X:  0x%08X (%2i bits)", (unsigned) address, (Bit32u) val, io_len * 8));
-
-#ifndef BX_LITTLE_ENDIAN
-  // the Guest will expect the return value to be in little-endian form
-  val = bx_bswap32(val);
-#endif
 
   return(val);
 }
@@ -1006,7 +1001,6 @@ bool bx_uhci_core_c::DoTransfer(Bit32u address, struct TD *td)
       case USB_TOKEN_OUT:
       case USB_TOKEN_SETUP:
         if (maxlen > 0) {
-          // no need to endian this. It is little endian no matter what
           DEV_MEM_READ_PHYSICAL_DMA(td->dword3, maxlen, p->packet.data);
         }
         ret = broadcast_packet(&p->packet);
@@ -1034,7 +1028,6 @@ bool bx_uhci_core_c::DoTransfer(Bit32u address, struct TD *td)
         ret = USB_RET_BABBLE;
       }
       if (len > 0) {
-        // no need to endian this. It is little endian no matter what
         DEV_MEM_WRITE_PHYSICAL_DMA(td->dword3, len, p->packet.data);
       }
     } else {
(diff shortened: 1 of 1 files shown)


Randrianasulu:
This does not compile on gcc15:

uhci_core.cc:732:66: error: cannot convert ‘Bit32u*’ {aka ‘unsigned int*’} to ‘Bit8u*’ {aka ‘unsigned char*’}         

uhci_core.cc:733:66: error: cannot convert ‘Bit32u*’ {aka ‘unsigned int*’} to ‘Bit8u*’ {aka ‘unsigned char*’}         

uhci_core.cc:763:50: error: ‘bit32u’ was not declared in this scope; did you mean ‘Bit32u’?          763 |       DEV_MEM_READ_PHYSICAL(address +  0, sizeof(bit32u), &td.dword0);

Randrianasulu:
Sadly, win98 still does not work with tablet connected to port1 on uhci controller with this mr manually applied.

It still works with same tablet  connected to ohci controller.

All on BE host (Linux, Bonslack ppc32 -current)

blin-livecd seems to have its own problems with usb mouse/tablet  autoconfiguration  disregard  it for now ...

fysnet:
The OHCI works on your big-endian machine, yes?

In the OHCI code, the following line is used to return the value of the read handler:
https://github.com/bochs-emu/Bochs/blob/1acc7f49c71652f95513ab388e8f14001fa90e63/bochs/iodev/usb/ohci_core.cc#L574
That doesn't seem to be Endian independent, yet it works, yes?

In the UHCI code, the following line is used to return the value of the read handler:
https://github.com/bochs-emu/Bochs/blob/1acc7f49c71652f95513ab388e8f14001fa90e63/bochs/iodev/usb/uhci_core.cc#L367
On a big-endian machine, the value is returned as big-endian.

What if you changed the line to:
```
  // yes, I know this isn't the correct way to do this, but I want to check something...
  return bx_bswap32(val);
```
Does the tablet then work?

(Make sure you still have this PR applied)

Also, just curious, have you tested a USB mouse device on both OHCI and UHCI? Does the mouse work on OHCI but not UHCI? I ask because is this problem a big-endian problem, or a `tablet` problem?

Randrianasulu:
It stranger and stranger.

On BE host, with patch applied
tablet works on both OHCI and UHCI
usb mouse does not work on both.

Moreover, usb  mouse behaves really strange on LE Android  tablet in win98se (tried both uhci and ohci)  - clicks are registered, but movement is way off.

I disabled mouse completely for those tests in .bochsrc. Or it need to be enabled even for usb mouse?

Tablet device seems to create its own hid mouse.

Without patch on BE host
tablet at uhci back to generating enormous log file and slowing down win98se guest to a crawl.

So I think patch is working, just usb  mouse behaviour is strange on both hosts, with my copy of win98se at least. Probably separate issue.

Does usb mouse work for you?

Randrianasulu:
Just reapplied patches, usb   tablet device works on boch ohci and uhci. (on BE).

On uhci it feels choppy, but works.

bochsout is also at 112 k so something probably still unhappy.

fysnet:
Stanislav, it looks like this PR with the additional `bx_bswap32(val)` will make the UHCI endianness aware. However, what would be the correct way of returning `val` as little-endian, no matter the host endianness?
```
  return bx_bswap32(val);  // this needs to be changed to????
```
We could use `WriteHostDWordToLittleEndian()` but that implies that we already know the Guest address to write the value to. Since this is in a read_handler, we don't know that yet.

I think all we need to do is return `val` as little-endian, no matter the host endianness.

Would the following be acceptable?
```
#ifdef BX_LITTLE_ENDIAN
  return(val);
#else
  return bx_bswap32(val);
#endif
```

Randrianasulu,

I tried the mouse (and tablet) with a Win98SE Guest and it is too, very abnormal. Both UHCI and OHCI.
The same mouse (and tablet) on a WinXP host acts normal.
I think it is Win98. I have always had Win98 be abnormal. It is something I should investigate to see why.
However, I don't believe it is anything to do with endianness.

With this PR and the additional change of `return bx_bswap32(val);`, can you say that the UHCI now works on a big-endian machine?

Randrianasulu:
with last commit applied win98se guest seems to hang with uhci + tablet .... (on BE)

https://github.com/fysnet/Bochs-USB-Edition/commit/667e98039bc6b7faad04d5aed1f5ad52cadf6da9.diff

stlintel:
> Stanislav, it looks like this PR with the additional `bx_bswap32(val)` will make the UHCI endianness aware. However, what would be the correct way of returning `val` as little-endian, no matter the host endianness?
> 
> ```
>   return bx_bswap32(val);  // this needs to be changed to????
> ```
> 
> We could use `WriteHostDWordToLittleEndian()` but that implies that we already know the Guest address to write the value to. Since this is in a read_handler, we don't know that yet.
> 
> I think all we need to do is return `val` as little-endian, no matter the host endianness.
> 
> Would the following be acceptable?
> 
> ```
> #ifdef BX_LITTLE_ENDIAN
>   return(val);
> #else
>   return bx_bswap32(val);
> #endif
> ```
> 
> Randrianasulu,
> 
> I tried the mouse (and tablet) with a Win98SE Guest and it is too, very abnormal. Both UHCI and OHCI. The same mouse (and tablet) on a WinXP host acts normal. I think it is Win98. I have always had Win98 be abnormal. It is something I should investigate to see why. However, I don't believe it is anything to do with endianness.
> 
> With this PR and the additional change of `return bx_bswap32(val);`, can you say that the UHCI now works on a big-endian machine?

I think may be just add 

#ifndef BX_LITTLE_ENDIAN
  val = bx_bswap32(val);
#endif

and add commentary why it is needed. Honestly for me BE is also very confusing and I didn't try to understand yet what would be the right way to handle it.

fysnet:
> On uhci it feels choppy, but works.

I remember now why Win98se doesn't handle the UHCI well. It never got the Toggle bit correct.

Set the following to zero
https://github.com/bochs-emu/Bochs/blob/b771a3dc1a520f1b717aadc1c0e62ca3057d527e/bochs/iodev/usb/usb_common.h#L157
This should make Win98 handle UHCI HID devices better.

Vort:
> with last commit applied win98se guest seems to hang with uhci + tablet .... (on BE)

> and add commentary why it is needed.

If this change breaks UHCI support, then it either not needed at all, or needed in combination with some other changes.

Randrianasulu:
> > On uhci it feels choppy, but works.
> 
> I remember now why Win98se doesn't handle the UHCI well. It never got the Toggle bit correct.
> 
> Set the following to zero
> 
> https://github.com/bochs-emu/Bochs/blob/b771a3dc1a520f1b717aadc1c0e62ca3057d527e/bochs/iodev/usb/usb_common.h#L157
> 
> 
> This should make Win98 handle UHCI HID devices better.

Yes, this fixes choppines in win98se with usb tablet hanging from uhci port (still using only 3 first patches from this PR).

But I wonder why next patch makes win98 hang?

fysnet:
> But I wonder why next patch makes win98 hang?

So that I am sure what you mean, please specify the modification that makes it hang. This way I make sure I know which patch you are talking about.  Also, does win98 work without the patch?

Randrianasulu:
> > But I wonder why next patch makes win98 hang?
> 
> So that I am sure what you mean, please specify the modification that makes it hang. This way I make sure I know which patch you are talking about. Also, does win98 work without the patch?

one modifying 
bx_uhci_core_c::read

I think I can only download patches via webinterface if I follow to their parent repo.

so 667e98039bc6b7faad04d5aed1f5ad52cadf6da9.diff dhould be in you repo / branch ?

This PR contain for nowfive commits, I applied first two chronologically, fixet bit vs Bit by hand and fixed uhci toggle as you suggested.

My understanding that last patch in this commit series just simplified fourth patch, and 4th patch was making my win98se hang, twice. (on BE host)

fysnet:
So if I remove that patch:
```
#ifdef BX_LITTLE_ENDIAN
  return(val);
#else
  return bx_bswap32(val);
#endif
```
and make it
```
  return(val);
```
again, reverting back to its original source, does Win98 work on your BE machine? i.e.: Was this last patch irrelevant?

Randrianasulu:
> So if I remove that patch:
> 
> ```
> #ifdef BX_LITTLE_ENDIAN
>   return(val);
> #else
>   return bx_bswap32(val);
> #endif
> ```
> 
> and make it
> 
> ```
>   return(val);
> ```
> 
> again, reverting back to its original source, does Win98 work on your BE machine? i.e.: Was this last patch irrelevant?

Sorry, I just not applied it  this time 

I did

git checkout bochs/iodev/usb/uhci_core.cc

and then applied 

cat bdd8f1861da5cd73971757b3b104601a9c3c67bf.diff | patch -p1

cat 5a8020d6b2e5d50d0ba18866047a4667c1116a8c.diff | patch -p1

and manually edited same place where you made Bit vs bit typo. And then your suggestion about usb toggle.

I can re-apply this patch again, and see if it hang again - but then what? How to debug this?

Randrianasulu:
re applied  4th commit, win98 predictably hanged . ;)

```
tail bochsout.txt
00122347782e[CPU0  ] interrupt(): soft_int && (gate.dpl < CPL)
00122348517e[CPU0  ] interrupt(): soft_int && (gate.dpl < CPL)
00122787930e[CPU0  ] interrupt(): soft_int && (gate.dpl < CPL)
00122788658e[CPU0  ] interrupt(): soft_int && (gate.dpl < CPL)
00122838314e[CPU0  ] interrupt(): soft_int && (gate.dpl < CPL)
00122839049e[CPU0  ] interrupt(): soft_int && (gate.dpl < CPL)
00124202447e[CPU0  ] RET_Op32_Iw: offset outside of CS limits
00124303078e[CPU0  ] branch_near32: offset outside of CS limits
00124466773e[CPU0  ] RET_Op32_Iw: offset outside of CS limits
00124564000e[UHCI  ] UHCI Core: ISO bit is set...
```

this is tail of bochsout as of now.

stlintel:
> re applied 4th commit, win98 predictably hanged . ;)

May be this 4th commit not required ?
How does it behave with all the rest without it ?

Randrianasulu:
> > re applied 4th commit, win98 predictably hanged . ;)
> 
> May be this 4th commit not required ? How does it behave with all the rest without it ?

On BE host win98se can use uhci tablet.

But breaking news! I left win98 in hanged state for some time and it booted ... into safe mode o.O No mouse still ... (this is witch 4th patch applied).

Should I copy bochsout.txt that comes out of booting win98se on BE host with first 3 patches applied? It can be few mbytes, so zip it?

stlintel:
> > > re applied 4th commit, win98 predictably hanged . ;)
> > 
> > 
> > May be this 4th commit not required ? How does it behave with all the rest without it ?
> 
> On BE host win98se can use uhci tablet.

So what the 4th patch needed for ?
What is not yet working with 1st 3 ?

vruppert:
After having a quick look at the code I can say that byte swapping in the read handler is wrong (registers are in dative endian and cpu expects this format). No other device uses byte swapping at this case and write handler would also be affected. The comments at the DMA functions are not necessary, since DMA is used for data blocks that don't need changes in byte ordering.
I have no access to a BE machine, but I think the other code is okay.

stlintel (approved):

PR #773, opened 2026-07-04
Fix Compilation on MacOS X

eviessh:
I run a lot of older hardware, basically any laptop I can get my hands on. In that set of laptops, I have two MacBooks, both running different versions (11.7 and 26), and compiling Bochs on either version was difficult at best, and completely manual at worst. I aim to improve that by providing infrastructure that builds on what's already here.

Attached to this pull request are a bunch of changes that I needed to add to get Bochs to compile on 11.7. Note that these are not particularly well tested, and I would _love_ some extra testing if anyone can provide. A summary of the changes are as follows;

- Added the ability to add a custom `ar`.
- Fixed the application of unsupported flags to Clang.
- Made the build script executable by default.
- Added the automatic ability to remove sound / networking.
  - This change is documented in the `.conf.macosx` file, but the gist is that for a LOT of MacOS versions, QuickTime doesn't exist for sound. I have also had problems with networking failing to compile on all MacOS machines I own, so I added an option to remove that from the build as well. I plan to investigate the source of that further.

These changes are quite a lot in terms of code-size. This little escapade was not supposed to go like this, lol. Please, if there is anything I need to improve, document, or change, please let me know. If you'd rather this pull request be split up, I can do that as well.


eviessh:
I figure I'll take a look at maintaining the Carbon backend next, if that help is still needed. I am very very new to this project, so any guidance would be extremely helpful. If there is a formatting issue with anything I've sent in, please let me know. Thank you guys for all your work.

vruppert:
Both Windows workflows are failing, so some more investigation is required.

stlintel:
Seems like most of the changes are softfloat3e, lets try to split it up. I also have comments about stuff you done there so lets discuss it within new PR

stlintel:
And btw, why don't you just compile softfloat as C++ instead for you?

Vort:
Few notes regarding softfloat3e:
* Older versions of clang also print lots of `warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated`. I just tested main branch with clang 18.1.8, which is 2 years old;
* These warnings are not specific to MacOS. I'm seeing them with Windows (MSYS2) and saw with Termux.

eviessh:
> Few notes regarding softfloat3e:
> 
>     * Older versions of clang also print lots of `warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated`. I just tested main branch with clang 18.1.8, which is 2 years old;
> 
>     * These warnings are not specific to MacOS. I'm seeing them with Windows (MSYS2) and saw with Termux.

@Vort Noted! I have no other environment available right now to test it on, and I saw no other mentions of this issue, so I figured it was MacOS-specific. I'll separate those changes out into a separate PR.

eviessh:
It seems that the failure to build for the Windows binary test is caused by me not setting a default for AR, which is an easy fix. However, the other failure is more mystifying and I will investigate further.

eviessh:
I'm having trouble figuring out why exactly the workflow is failing--in all my testing, even if AR is not set, a suitable replacement is found. This is very likely a case of me missing something silly, but if anyone could offer some pointers that'd be great. In the meantime, I'll keep looking.

Vort:
I was able to reproduce problem with empty `AR` variable with MSYS2.
However, I don't know how to debug such problems.

> I have no other environment available right now to test it on

MSYS2 can be installed to Windows 8.1 virtual machine for example.

But there will be needed some changes to `.conf.win32-cygwin` for build to work:
* `CC="clang"`
* `CXX="clang++"`
* `--disable-readline \`

It is also possible to make tests with cross build from Linux, like it is done in GitHub Actions.

Vort:
Shouldn't there be `autoreconf` pass after such change?
Can't make it to work even with regenerated `configure` however:
`AR=''` still appears in `config.log`.
I was able to change it only with `AR=ar ./.conf.win32-cygwin` command, but this is not how it should be done I suppose.

Vort:
45ea2b6359eb3c91c6032b16fe240af816348538 builds successfully for me with old MSYS2.
Newer MSYS2 fails with `error: invalid argument '-std=gnu23' not allowed with 'C++'` on this line:
https://github.com/bochs-emu/Bochs/blob/e1ddf6b9f77e8906b65bcc314bdc39c8170696e8/bochs/Makefile.in#L289
I assume it behaves like that because `configure` was regenerated with newer Autoconf.

eviessh:
Yes, that line was originally corrected by me in the pull request, but I removed that change to the Fix Softfloat Treatment PR (#774) because I felt it fit that bill better. That error is failing compilation for me as well on Clang 21.1.0, but it's just a simple typo and the fix is two characters, lol. I can just pull that change back to this branch, if you think that would be beneficial.

Vort:
> I can just pull that change back to this branch, if you think that would be beneficial.

Yes, I think it should be done:
1. It is not related to softfloat;
2. It is a regression.

Regarding #774, can you confirm if my suggestion with renaming of `.c` -> `.cc` works?
If so, it may be better to close #774 and reopen new one with rename approach.
Regarding `build\win32\vs2019-workspace\*` changes - I guess it will be enough to make similar replacements in `softfloat3e.vcxproj`. Can be done "blindly" if you are not going to install MSVC and checked later either by me or by @vruppert.

eviessh:
Added that change from over on that branch, and I agree. I'll close #774 and reopen a new PR.

Vort:
I tested 129fbe66a7ec3bd972901b8852e6d836c9b862c6 with Windows (2 MSYS2 and MSVC2019) and Termux - all builds were fine.

stlintel (approved):

PR #774, opened 2026-07-04
Fix Softfloat Treatment as C++

eviessh:
Softfloat is being compiled under C++ mode, which is deprecated in Clang and soon to be removed. This patchset simply updates all required code to allow softfloat and all code that depends on it to be compiled properly and without deprecated behavior.

Any testing is 100% welcome, and if there are any changes I have to make I will gladly make them.


stlintel:
It fails CI/DI all over...
I still suggest to compile softfloat as C++ code instead.

Vort:
> I still suggest to compile softfloat as C++ code instead.

Isn't this what is causing compilation problems?
Can you clarify your idea more?

For me it looks like clang compiler suggests to separate C and C++ parts better, which is a good idea overall, because these languages are different despite being similar.
But it should be done properly of course.

stlintel:
The entire softfloat stuff is actually C++
when we compile it I even use g++ and not gcc:
image

Did you try to tell Clang to compile it all as C++ ignoring ".c" file extension and see if it goes smoothly ?

Vort:
I tried slightly different thing:
1. Renamed all `*.c` files in `softfloat3e` directory to `*.cc`;
2. Made 6 corresponding changes to `Makefile.in`.

No more warnings appeared during compilation with older MSYS2.

If such solution is acceptable, I prefer it more than trying to preserve C specifics.
It will require corrections to MSVC project files however.

stlintel:
I think it is acceptable

eviessh:
Sorry for the late contribution to the discussion, but I think the failure is because of how Windows handles C vs. C++ code. Name mangling is completely different between the two, and MSVC is just throwing a bunch of "undefined symbol" because it can't understand the difference without guidance. I will try to fix it if you think that's a good path forward, but that will likely take a while for me to get a suitable testing environment up and figure out the changes needed. If you think the better route is just making softfloat3e a C++ module like the rest of the project, I can go along with that as well.

Regardless, just let me know. Thanks!

stlintel:
>If you think the better route is just making softfloat3e a C++ module like the rest of the project, I can go along with that as well

Yes, I think this would be simpler and less intrusive

eviessh:
Going to close this PR then and reopen a new one. :)

eviessh:
If anyone's interested in what the problem was, I just found it. Turns out that MS has an option to "compile code as ___" in their build files, and for softfloat it was still set up under C++, which was causing the "undefined symbol" errors. Rookie mistake on my part! Lol.

PR #778, opened 2026-07-07
Fix -Werror Configuration on Linux

eviessh:
I'm trying to sort through some of the warnings I get when compiling on various platforms, and specifically on Linux I noticed that configuring at all was impossible, as the configuration script couldn't find the POSIX thread library. I found that the main problem on this front was that the POSIX check script was malformed in several ways, and fixed them with no change to the actual underlying functionality.

### DIFF
--- a/bochs/.conf.linux
+++ b/bochs/.conf.linux
@@ -2,6 +2,7 @@
 #
 # .conf.linux
 #
+
 if test $# = 0; then
   which_config=plugins
 else
@@ -18,7 +19,6 @@ export CC
 export CXX
 export CFLAGS
 export CXXFLAGS
-
 
 case $which_config in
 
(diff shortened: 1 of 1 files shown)


sskras:
Ping @stlintel, @vruppert:

is it possible that `bochs/configure.ac` is copied from the Autotools, the upstream project for this file, which potentially might have fixed this issue in the meantime?

If so, maybe it would be worth updating the related files from newer (or even latest) Autotools version?

Vort:
> is copied from the Autotools, the upstream project for this file

Can you specify where exactly it can be copied from?

bachimusprime:
autotools isn't the upstream for that file, it needs to be different for every autotools project lol

eviessh:
@sskras Autotools is a build system, and `configure.ac` is the configuration file for that build system. If you know CMake or Meson, this is a parallel for their `CMakeLists.txt` and `meson.build` respectively. :)

sskras:
Ah, OK – I made a blind shot, and mixed things up in my mind.

It's `aclocal.m4` which is generated by `aclocal` coming from the Autoconf project.

And `configure.ac` is the hand-written input file:
https://opensource.com/article/19/7/introduction-gnu-autotools#autotools-configuration
https://tomlee.co/2012/08/autotools-for-humans-part-1/#configureac
https://coin-or-tools.github.io/BuildTools/configure.html#the-beginning-of-a-configureac-file

So seemingly it's `aclocal` and `autoconf` executables from the Autoconf project that need to be updated from time to time, IIUC now.

Of course, there is also the `autoscan` tool which can generate `configure.ac` automatically:
https://www.baeldung.com/linux/configure-script-creation#1-generate-a-configureac-template-file
... but it output merely an initial template, nothing project-specific, it seems.

So again please excuse me for the confusion.

eviessh:
Absolutely nothing wrong with a bit of a mix-up, lol. Autotools is a confusing build system at best.

And, to maintainers, so sorry for the re-trigger of workflows! Accidentally pushed a commit to this branch that was bound for another and had to reset the branch, sorry!

vruppert:
The changes are okay so far, but using the "allwarnings" case doesn't make sense to me. It produces errors in pieces of code we never want to change (like the stubs in paramtree.h and siminterface.h). Generating all possible warnings is okay, but when stopping compilation at the first file (devices.cc) you'll never see all warnings.

stlintel:
I had the same comment. The configure fix seems important, all-warnings seems like misplaced ...

eviessh:
Yes, you're right, now that I think about it. I'll fix that.

vruppert (approved):

Issue #811, opened 2026-07-27
Broken build with disabled SMF support


Vort:
When I run `./.conf.win32-cygwin`, change `#define BX_USE_CIRRUS_SMF` to `0` in `config.h` and run `make -j 3`, compilation errors appear:

<details><summary>Errors list</summary>
<p>

```
svga_cirrus.cc:1461:3: error: too few arguments to function call, expected 4, have 3; did you mean '::write'?
 1461 |   VGA_WRITE(address,value,io_len);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
svga_cirrus.cc:61:48: note: expanded from macro 'VGA_WRITE'
   61 | #define VGA_WRITE(addr,val,len)  bx_vgacore_c::write(addr,val,len)
      |                                                ^~~~~
D:/msys64/clang64/include/io.h:346:15: note: '::write' declared here
  346 |   int __cdecl write(int _Filehandle,const void *_Buf,unsigned int _MaxCharCount) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
      |               ^
svga_cirrus.cc:2324:5: error: too few arguments to function call, expected 4, have 3; did you mean '::write'?
 2324 |     VGA_WRITE(address,value,1);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~
svga_cirrus.cc:61:48: note: expanded from macro 'VGA_WRITE'
   61 | #define VGA_WRITE(addr,val,len)  bx_vgacore_c::write(addr,val,len)
      |                                                ^~~~~
D:/msys64/clang64/include/io.h:346:15: note: '::write' declared here
  346 |   int __cdecl write(int _Filehandle,const void *_Buf,unsigned int _MaxCharCount) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
      |               ^
svga_cirrus.cc:2554:5: error: too few arguments to function call, expected 4, have 3; did you mean '::write'?
 2554 |     VGA_WRITE(address,value,1);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~
svga_cirrus.cc:61:48: note: expanded from macro 'VGA_WRITE'
   61 | #define VGA_WRITE(addr,val,len)  bx_vgacore_c::write(addr,val,len)
      |                                                ^~~~~
D:/msys64/clang64/include/io.h:346:15: note: '::write' declared here
[...]

PR #812, opened 2026-07-28
Implement method handlers for GeForce

Vort:
This PR properly fixes `compiler limit: blocks nested too deeply` error from Visual Studio.
There should be no functional changes.
Related issue: #811.

### DIFF
--- a/bochs/iodev/display/geforce.cc
+++ b/bochs/iodev/display/geforce.cc
@@ -191,6 +191,7 @@ bool bx_geforce_c::init_vga_extension(void)
   BX_GEFORCE_THIS svga_init_members();
   BX_GEFORCE_THIS svga_init_pcihandlers();
   BX_GEFORCE_THIS bitblt_init();
+  BX_GEFORCE_THIS init_method_handlers();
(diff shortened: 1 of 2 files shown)


stlintel (approved):

bochs/iodev/display/geforce.cc
```diff
@@ -6120,940 +6121,1568 @@ void bx_geforce_c::execute_sifm(gf_channel* ch, Bit32u cls, Bit32u method, Bit32
...
+{
+  ch->d3d_semaphore_obj = param;
+}
+
+void d3d_mh_report_obj(bx_geforce_c* this_ptr, gf_channel* ch, Bit32u cls, Bit32u method, Bit32u param)
+{
+  ch->d3d_report_obj = param;
+}
+
+void d3d_mh_clip_horizontal(bx_geforce_c* this_ptr, gf_channel* ch, Bit32u cls, Bit32u method, Bit32u param)
```
stlintel, line 6271: what the difference between these C-like methods and C++ class members ?
Why not to make them class members and use SMF trick to make them static members if SMF=1?

PR #816, opened 2026-07-31
Add SMF 0 support for GeForce method handlers

Vort:
This change fixes build with `#define BX_USE_GEFORCE_SMF  0`.
There should be no functional changes.


stlintel:
I think for simplicity it is better to merge it without fixing logio stuff and let Volker to apply fix later
Can you uncomment all logio stuff and remove fixme comments ?

stlintel (approved):

Vort:
> Can you uncomment all logio stuff and remove fixme comments ?

Ok, done.

PR #826, opened 2026-08-16
Attempt to implement #FERR and #IGNNE handling as well as MSDOS compatibility mode of handling x87 unmasked exceptions

stlintel:
Attempt to implement #FERR and #IGNNE handling as well as MSDOS compatibility mode of handling x87 unmasked exceptions

Untested, should run extensive testing to see if x87 unmasked exceptions delivered properly.

I guess then adding a Bochs option to control this behavior will be the must:
extfpuirq: mode=...

And add deprecated FERR# and IGNNE#

### DIFF
--- a/bochs/CHANGES
+++ b/bochs/CHANGES
@@ -2,6 +2,7 @@ Changes in 3.1 (coming soon):
 
 Brief summary :
   - CPU: Implemented FRED, USER_MSR, MOVRS, AMX-FP8 ISA extensions emulation
+  - CPU: Implemented MS-DOS Compatibility mode for handling x87 FPU exceptions, FERR# and IGNNE# pins enabled when 'extfpuirq' plugin is enabled
   - CPUDB: Added new Wildcat Lake CPU definition featuring FRED, WRMSRNS and MSRLIST
   - Bugfixes for CPU emulation correctness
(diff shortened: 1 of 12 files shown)


stlintel:
Summary of changes:

- Implement IGNNE# as pc_system bit, CPU can read it directly from there when needed
- Implement raising and clearing FERR# by CPU (done using DEV_extfpuirq_set_fpu_error interface)
- When CPU executes 'x87 waiting' instruction in legacy mode + FCW.ES is set (indicating FERR#) and IGNNE# is 0
  Send CPU to HLT like state until NMI, INTR or other event occur or IGNNE# is raised

vruppert:
I can confirm that the code can be built for Linux and Windows. I don't have a test application here, so I intend to wait for the test results from @lab-bat. I'm not an FPU expert, but the other code looks okay to me.
I don't think that we need a config option for it, since there are basically 3 situations:
1. ISA-only system: The external logic is always present (enabled in extfpuirq init())
2. PCI with PIIX/PIIX3/PIIX4 ISA bridge: The coprocessor error bit enables extfpuirq functionality (done by BIOS)
3. Newer PCI/PCIe chipsets: The control bit to enable this feature no longer exists

In addition to this the extfpuirq logic is an optional plugin. If it is not loaded the FERR# signal is silently ignored.

vruppert (dismissed):

lab-bat:
The result is correct.

PR #828, opened 2026-08-19
i486dx4: expose CR4 to support VME/PVI

Yggdrasill:
Hi,

Just another reasonably small patch, because I ran into a bit of a problem while writing some code for CR4 detection.

**Description**: Commit f30cdc367 introduces emulation of the i486DX4 CPU which is supported when compiled with `--enable-cpu-level=4`, but the commit does not expose the `MOV_RdCR4`/`MOV_CR4Rd` instructions to the decoder with the `BX_ISA_486` feature. I think I understand why, because only the later i486 CPUs like the DX4 actually has the CR4 register. However, this leaves VME/PVI broken when emulating these CPUs, as the only way to enable them is through the CR4 register. The assembly code that breaks is:

```as
check_cpuid:
    xor   eax, eax
    pusha
    pushfd
    pushfd
    pushfd
    btr   dword [esp], 21
    popfd
    pushfd
    pop   eax
    bts   dword [esp], 21
    popfd
    pushfd
    pop   ecx
    popfd
    xor   ecx, eax
    popa
    setnz al
    ret

check_cr4:
    pusha
    call  check_cpuid
    jz    check_cr4_ret
    cpuid
    ; CPUID leaf 1 test for:
    ; CR4.MCE, CR4.PAE, CR4.TSD,
    ; CR4.PSE, CR4.DE, CR4.VME
    ; Definitely got CR4 if non-zero.
    test  dl, 0xDE
check_cr4_ret:
    popa
    ret
```

This patch fixes that by changing the related lines in `cpu/decoder/ia_opcodes.def` to make the instructions valid for `BX_ISA_486`. It also corrects compilation with `--enable-cpu-level=4` by moving some `BX_CPU_LEVEL >= 5` guards so that features limited to these CPUs are correctly guarded, and lowers others to `BX_CPU_LEVEL >= 4` where appropriate. There is also an additional small fix, where a CPU level 6 guard has been split up. Specifically, it fixes that a `BX_CPU_LEVEL >= 6` guard encloses `BX_CPU_C::get_efer_allow_mask()` which makes a build with `--enable-cpu-level=5` fail. This function is now guarded with `BX_CPU_LEVEL >= 5`, followed again by a guard for CPU level 6.

The Intel i486 manual documents the existence of CR4 and its defined bits in Figure 4-6, p. 4-17. It also documents the PSE bit as present, but I cannot find any real i486 architecture CPUs that actually support this bit. The manual qualifies that usage of CR4 and the features controlled by it must first be checked by the CPUID instruction.

Please correct me if I am wrong about this, but: In terms of SMM I have simply retained the `SMRAM_FIELD_CR4` field, which I think should be sufficient for emulation purposes. The same manual documents that CR4 is saved in the SMM save state, but I have not been able to find public documentation of its actual location. The i486 manual simply says it is within a reserved region, though the existing mapping from 0x7F14 does exist within a reserved region in the i486 SMRAM state save map on page 8-9.

**Testing**: I have done manual testing under both `--enable-cpu-level=4` and `--enable-cpu-level=5`, and writes of bit 0 and 1 on the i486dx4 model behave correctly, and any higher bits cause the CPU to fault as intended. The i386 model still produces #UD exceptions when trying to access CR4, and CPU level >=5 models also behave correctly with setting/unsetting other bits, e.g. PSE etc.

I will say one thing, which is that it would possibly be better if the `BX_ISA_486` feature was not used for enabling these instructions in `cpu/decoder/ia_opcodes.def`, for the aforementioned reason that not all i486 CPUs actually have CR4. I don't know how appropriate it would be to define something like a `BX_ISA_CR4` feature, because on its own its not really an ISA like modelled by `BX_ISA_SVM` etc. I'd be very open to changing that if requested, because this patch is obviously not generally correct for all future level 4 CPUs.

Regards,
Yggdrasill

### DIFF
--- a/bochs/bx_debug/dbg_main.cc
+++ b/bochs/bx_debug/dbg_main.cc
@@ -1121,9 +1121,12 @@ void bx_dbg_info_control_regs_command(unsigned cpu)
   dbg_printf("    PCD=page-level cache disable=%d\n", (cr3>>4) & 1);
   dbg_printf("    PWT=page-level write-through=%d\n", (cr3>>3) & 1);
 
-#if BX_CPU_LEVEL >= 5
+#if BX_CPU_LEVEL >= 4
   Bit32u cr4 = SIM->get_param_num("CR4", dbg_cpu_list[cpu])->get();
   dbg_printf("CR4=0x%08x: %s\n", cr4, stringify_CR4(cr4, s));
+#endif
+
+#if BX_CPU_LEVEL >= 5
 #if BX_SUPPORT_X86_64
   if (BX_CPU(cpu)->is_cpu_extension_supported(BX_ISA_LONG_MODE)) {
     dbg_printf("CR8: 0x%x\n", BX_CPU(cpu)->get_cr8());
--- a/bochs/cpu/cpu.cc
+++ b/bochs/cpu/cpu.cc
@@ -638,7 +638,7 @@ void BX_CPU_C::prefetch(void)
 #endif
   {
 
-#if BX_CPU_LEVEL >= 5
+#if BX_CPU_LEVEL >= 4
     if (USER_PL && BX_CPU_THIS_PTR get_VIP() && BX_CPU_THIS_PTR get_VIF()) {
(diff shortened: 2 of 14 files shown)


stlintel (approved):

stlintel:
Approvee

PR #835, opened 2026-08-25
LFN to SFN fix

### LINKED ISSUE #832: Incomplete LFN support with vvfat
When vvfat disk is created from files unpacked from attached archive, only `LongNameA.txt` .. `LongNameJ.txt` files are displayed correctly in Windows 98 guest. `LongNameK.txt` is displayed as `Longna00.txt` instead.

Image

Test files: [LongNames.zip](https://github.com/user-attachments/files/31350481/LongNames.zip)
Version: acdc1da

fysnet:
Try to fix https://github.com/bochs-emu/Bochs/issues/832


bochs/iodev/hdimage/vvfat.cc
```diff
@@ -552,14 +552,66 @@ void vvfat_image_t::init_fat(void)
...
+    memset(sfn, ' ', 11);
+    memcpy(sfn, base, strlen(base));
+    memcpy(sfn + 8, ext, strlen(ext));
+    sfn[11] = '\0';
+    return;
+  }
+  
+  // Append short tail designation
+  // (crashes if sequence > 9999999 ?)
+  sprintf(sequ, "~%i", sequence);
```
stlintel: if gave only 5 bytes for sequ, probably can explain why crashes if sequence > 9999999
with such small buffer it would overflow when sequence > 9999

stlintel (changes requested):

bochs/iodev/hdimage/vvfat.cc
```diff
@@ -552,14 +552,66 @@ void vvfat_image_t::init_fat(void)
...
+  base[max] = '\0';
+  strcat(base, sequ);
+  
+  // Format as 8.3 string
+  sprintf(sfn, "%-8s%-3s", base, ext);
+  
+  // Remove space padding if needed or keep fixed 11 bytes
+  for (int k=0; k<11; k++) {
+    if (sfn[k] == ' ')
+      sfn[k] = ' ';
```
Vort: If it's equal to space then it's equal to space.
What does it mean? Do I miss something?

bochs/iodev/hdimage/vvfat.cc
fysnet: You're right, but the crash I was thinking more of is if the length of `sequ` becomes more than 8, the `int max = 8 - (int) strlen(sequ);` would now make it crash.  The crash you suggest is an easy fix. The crash I am thinking about is not. :-)

bochs/iodev/hdimage/vvfat.cc
fysnet: Sorry, this is a remnant of the debugging phase. I have removed it. After a few more modifications, I will update the code.

fysnet:
The non-Win32 version requires BX_MAX_PATH.
(Sorry about the addition/deletion of vvfat_old.cc. It was selected by mistake)

bochs/iodev/hdimage/vvfat.cc
```diff
@@ -552,14 +552,94 @@ void vvfat_image_t::init_fat(void)
...
+  // we need to copy the filename to a temp location so we don't modify the passed name
+  strncpy(filename, lfn, BX_MAX_PATH-1);
+  filename[BX_MAX_PATH-1] = '\0';
+
+  // we need to change any illegal char to a legal char.
+  // a valid SFN char is:
+  //  Uppercase letters A-Z
+  //  Numbers 0-9
+  //  space char (though it must be trailing?)
+  //  one of the chars listed in legalchars[] above
```
Vort: Outdated comment?
I don't see mentions of `legalchars` anywhere else.

bochs/iodev/hdimage/vvfat.cc
Vort: I don't think it makes sense to develop perfect solution for such unlikely case. At least for now.

bochs/iodev/hdimage/vvfat.cc
fysnet: Correct, yes.

fysnet:
Okay, try this one please.
I agree that for the sake of Bochs, it only needs to return a unique filename, whether that name is completely accurate or not.
However, because I want it to be as accurate as possible...
Thanks,
Ben

bochs/iodev/hdimage/vvfat.cc
```diff
@@ -552,14 +552,120 @@ void vvfat_image_t::init_fat(void)
...
+  // do we need to add the '~' stuff?
+  if (!force && (base_len <= 8) && (ext_len <= 3)) {
+    memset(sfn, ' ', 11);
+    memcpy(sfn, base, strlen(base));
+    memcpy(sfn + 8, ext, strlen(ext));
+    sfn[11] = '\0';
+  } else {
+    // Append short tail designation
+    // (crashes if sequence > 999999 ?)
+    if (sequence > 999999)
```
stlintel, line 646: I suggest instead of PANIC return false from the function and handle it on caller side.
It should tell smth like this directly has too many files and abort gracefully.
Here is user choose continue after PANIC - nothing good gonna happen but also you cannot replace PANIC with FATAL because FATAL just would close Bochs at the end ...

stlintel (changes requested):

Vort:
> Okay, try this one please.

In my opinion, result looks good enough now, thanks.

bochs/iodev/hdimage/vvfat.cc
fysnet: Hi Stanislav,

You are right, it would be better to return and let the caller handle the error. However, the way the current code is written, the caller assumes the return will always be a valid entry pointer.

The code at `create_short_and_long_name()` assumes it will return a valid pointer every time.
The parent function that calls `create_short_and_long_name()`, `vvfat_image_t::read_directory()`, assumes a successful return.

There is nowhere within the code to handle a failed return.
To be able to handle the failed return, the whole file will need to be reworked to handle the failed return. Should this happen? Yes, of course.

However, since it is so unlikely that the PANIC will ever happen, I don't think it is worth the effort, at this point in time.

vruppert:
The long file names are looking okay in Explorer, but when I run scandisk (My Computer -> right click on drive symbol -> Properties -> Tools -> Check Error Status) in Win98 it still complains about the first appearance of a long file name.
This is my test code to create the files.
```
#include <stdio.h>

int main(int argc, char **argv)
{
  char fn[18];
  FILE *fd;

  for (int i = 0; i <= 100; i++) {
    sprintf(fn, "LongFilename %03d.txt", i);
    fd = fopen(fn, "w");
    fprintf(fd, "Test %03d\r\n", i);
    fclose(fd);
    fprintf(stderr, "%s\n", fn);
  }
  return 0;
}
```

fysnet:
A simple fix is to change Line 668 from:
`int sequence = 0;`
to
`int sequence = 1;`
The sequence number must start at 1, not zero.

However, looking through the code, the LFN's CRC field is calculated before the remainder of the SFN entry is created. Therefore, I have added that change as well, so that it is more accurate.
(I am wondering why Win98 didn't check for that)

vruppert:
First of all, now there is an unused variable warning at compilation time (long_index).
Win98 scandisk now complains about the 20th entry of the same name in the root folder (here: LongFilename 019.txt / LONGF~20.TXT). If there are only 19 items in the root, it depends on the disk size whether or not subfolders can have issues (504 MB yes / 1,5 GB no). I made all tests with FAT16 (< 2 GB), since FAT32 may have additional issues.

fysnet:
Please share how you are declaring the 'drive'. I use:
`ata0-slave: type=disk, mode=vvfat, path=vvfat, journal=vvfat.redolog`
and am not getting any errors:

image

fysnet:
However, in WinXP, the 32nd file is now in error:
image

image

`LongFilename 031.txt` is now missing, though shown as a SFN only.

Update: Try the new PR which fixes this specific issue.

fysnet:
One other issue I found is that if the Mouse is currently being captured and vvfat pops up the dialog to ask if you wish to commit any changes, the mouse is unavailable to click yes or no.

The mouse needs to be released before this dialog appears.

vruppert:
The mouse issue in the shutdown dialog and possible issues with FAT32 and scandisk I intend to investigate later. Now the VVFAT test disk with lots of similar file names works fine, but most of my existing VVFAT "drives" make Bochs crash at startup while reading the directory tree.

fysnet:
If the vvfat volume is a FAT32 volume, the current code is in error.

For example, the `.` and `..` entries do not contain a valid FAT32 entry.

Also, the current code may create date and times that use future dates and times which will make Win98 give an error.

image

I am working on this and will provide an updated PR to fix the erroneous dot and dotdot entries. The future date and time issue is really out of our control and I don't really think it needs to be fixed.

For example, unchecking the "invalid dates and times" checkbox
image
will bypass these future date tests.

fysnet:
> ... but most of my existing VVFAT "drives" make Bochs crash at startup while reading the directory tree.

At the following code:
https://github.com/bochs-emu/Bochs/blob/ff17a0c2bbabccf96d33af4e08ba8061889b079d/bochs/iodev/hdimage/vvfat.cc#L698-L704
Remove lines 702 and 703 and add a `break;` instead.

Does this fix the crash you are having?

If you are using WIN32, do this here:
https://github.com/bochs-emu/Bochs/blob/ff17a0c2bbabccf96d33af4e08ba8061889b079d/bochs/iodev/hdimage/vvfat.cc#L812-L818

Also, should we change the `BX_ERROR` to a `BX_PANIC`? Continuing after the Panic will be just fine as long as the `break` is added as explained above.

vruppert:
First of all, I can confirm that the scandisk date and time check is fixed now.
Unfortunately the hack from above does not fix the segfault I'm getting with several VVFAT drives. I think I have to add BX_INFOs here and there to find out what's going on.

**Update:**
Changing this in line 794 fixes the segfault:
```direntry_t *entry_long = NULL;```

Some other issues may exist, but with this change the PR can be merged.

vruppert (approved):

stlintel (approved):

PR #855, opened 2026-08-29
Fix equal-magnitude MIN_ABS/MAX_ABS selection in VRANGE instructions

ergrelet:
## Problem

The shared `f32_range` and `f64_range` helpers implement the `MIN_ABS` and
`MAX_ABS` operations used by `VRANGEPD`, `VRANGEPS`, `VRANGESD`, and
`VRANGESS`.

For absolute-value comparisons, the helpers clear both operand signs and use
a strict less-than comparison. When the operands have equal magnitudes but
opposite signs, equality therefore falls through to a source-order-dependent
choice: `MIN_ABS` selects SRC2 and `MAX_ABS` selects SRC1.

Consequently, one source order returns the positive operand for `MIN_ABS` and
the negative operand for `MAX_ABS`. This also affects `-0.0`/`+0.0` ties.

## Fix

Extend the existing opposite-sign selection path to cover equal-magnitude
absolute comparisons. The helpers compare magnitudes by clearing the IEEE-754
sign bit from each operand before testing for equality.

This follows the existing Bochs SoftFloat convention used by the float32,
float64, and float16 MINMAX helpers, as well as by the surrounding RANGE
code.

When the magnitudes are equal and the signs differ, the existing selection
logic then chooses the negative operand for MIN_ABS and the positive operand
for MAX_ABS, independent of source order. NaN handling and DAZ normalization
remain unchanged and occur before this selection.

## Architectural reference

The Intel® 64 and IA-32 Architectures Software Developer’s Manual, Volume 2,
sections “VRANGEPD—Range Restriction Calculation for Packed Pairs of Float64
Values,” “VRANGEPS—Range Restriction Calculation for Packed Pairs of Float32
Values,” “VRANGESD—Range Restriction Calculation From a Pair of Scalar
Float64 Values,” and “VRANGESS—Range Restriction Calculation From a Pair of
Scalar Float32 Values,” define the range operations in Tables 5-22 and 5-23.

For equal-magnitude operands with opposite signs, those tables require
`MIN_ABS` to select the negative operand and `MAX_ABS` to select the positive
operand.

- [Intel® Software Developer Manuals download page](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html)

## Validation

A standalone 64-bit CPL3 guest exercised the four instructions on the Bochs
`tigerlake` CPU model with AVX-512 state enabled (`XCR0=0xe7`). The tie tests
used both source orders for `-1.0`/`+1.0` and `-0.0`/`+0.0`, with sign control
set to preserve the selected operand’s sign.

| Instruction | Operations and tie inputs | Before | After |
|---|---|---:|---:|
| `VRANGEPD` | `MIN_ABS`/`MAX_ABS`, `±1.0` and `±0.0`, both orders | FAIL | PASS |
| `VRANGEPS` | `MIN_ABS`/`MAX_ABS`, `±1.0` and `±0.0`, both orders | FAIL | PASS |
| `VRANGESD` | `MIN_ABS`/`MAX_ABS`, `±1.0` and `±0.0`, both orders | FAIL | PASS |
| `VRANGESS` | `MIN_ABS`/`MAX_ABS`, `±1.0` and `±0.0`, both orders | FAIL | PASS |

Additional controls passed after the change:

- unequal-magnitude `MIN_ABS` and `MAX_ABS` inputs;
- immediate `0x02` source-1 sign control;
- packed merge masking;
- scalar active, merge-masked, and zero-masked forms;
- scalar upper-element preservation and ZMM upper-bit clearing; and
- quiet NaN, signaling NaN, DAZ, denormal-input, and exception-flag helper
  checks for both float32 and float64.

The patched tree was rebuilt successfully with `make -j2`.

### DIFF
--- a/bochs/cpu/softfloat3e/f32_range.cc
+++ b/bochs/cpu/softfloat3e/f32_range.cc
@@ -96,7 +96,8 @@ float32 f32_range(float32 a, float32 b, bool is_max, bool is_abs, int sign_ctrl,
     else if (aIsNaN) {
         z = b;
     }
-    else if (signA != signB && ! is_abs) {
+    else if (signA != signB && (! is_abs ||
+             ((a & ~0x80000000) == (b & ~0x80000000)))) {
         if (! is_max) {
             z = signA ? a : b;
         } else {
--- a/bochs/cpu/softfloat3e/f64_range.cc
+++ b/bochs/cpu/softfloat3e/f64_range.cc
@@ -96,7 +96,9 @@ float64 f64_range(float64 a, float64 b, bool is_max, bool is_abs, int sign_ctrl,
     else if (aIsNaN) {
         z = b;
     }
-    else if (signA != signB && ! is_abs) {
+    else if (signA != signB && (! is_abs ||
+             ((a & ~UINT64_C(0x8000000000000000)) ==
+              (b & ~UINT64_C(0x8000000000000000))))) {
         if (! is_max) {
             z = signA ? a : b;
         } else {


stlintel:
The AI messed up with the description, can it use less water and just give examples of inputs that currently produce incorrect results ?

Also you can ask the AI to write test program too.

And I would appreciate human answer. Because I already had hard time to read AI slope in your previous reports. In nearly all cases I approved it because I understood the code fix and not because AI slope was readable.

bochs/cpu/softfloat3e/f32_range.cc
```diff
@@ -96,7 +96,8 @@ float32 f32_range(float32 a, float32 b, bool is_max, bool is_abs, int sign_ctrl,
     else if (aIsNaN) {
         z = b;
     }
-    else if (signA != signB && ! is_abs) {
+    else if (signA != signB && (! is_abs ||
```
stlintel, line 99: "For equal-magnitude operands with opposite signs, those tables require
MIN_ABS to select the negative operand and MAX_ABS to select the positive
operand."

So the issue is only about _ABS forms ?
But this code is explicitly says it is NOT handling abs form.
ABS form handled in next if and it takes care of the condition of sign as well.
My claim - this is fix is bogus completely until counter-example is provided

stlintel (changes requested):

stlintel:
If I understood the description correctly - the issue is ONLY when 2 arg are same magnitude and different sign.
Bochs doesn't obey the table from SDM:

image

Correct ?

If so - try the following patch instead, it will be less intrusive (and therefore more correct) fix and also doesn't forget about float16:

```
diff --git a/bochs/cpu/softfloat3e/f16_range.cc b/bochs/cpu/softfloat3e/f16_range.cc
index 25bb1ccdd..ce36375a9 100644
--- a/bochs/cpu/softfloat3e/f16_range.cc
+++ b/bochs/cpu/softfloat3e/f16_range.cc
@@ -107,7 +107,8 @@ float16 f16_range(float16 a, float16 b, bool is_max, bool is_abs, int sign_ctrl,
         if (is_abs) {
             tmp_a = tmp_a & ~0x8000; // clear the sign bit
             tmp_b = tmp_b & ~0x8000;
-            signA = 0;
+            if (tmp_a != tmp_b)
+              signA = 0;
         }
 
         if (! is_max) {
diff --git a/bochs/cpu/softfloat3e/f32_range.cc b/bochs/cpu/softfloat3e/f32_range.cc
index bb1af93ff..1983ba444 100644
--- a/bochs/cpu/softfloat3e/f32_range.cc
+++ b/bochs/cpu/softfloat3e/f32_range.cc
@@ -107,7 +107,8 @@ float32 f32_range(float32 a, float32 b, bool is_max, bool is_abs, int sign_ctrl,
         if (is_abs) {
             tmp_a = tmp_a & ~0x80000000; // clear the sign bit
             tmp_b = tmp_b & ~0x80000000;
-            signA = 0;
+            if (tmp_a != tmp_b)
+              signA = 0;
         }
         if (! is_max) {
             z = (signA ^ (tmp_a < tmp_b)) ? a : b;
diff --git a/bochs/cpu/softfloat3e/f64_range.cc b/bochs/cpu/softfloat3e/f64_range.cc
index 1025f474d..64c06e4c1 100644
--- a/bochs/cpu/softfloat3e/f64_range.cc
+++ b/bochs/cpu/softfloat3e/f64_range.cc
@@ -107,7 +107,8 @@ float64 f64_range(float64 a, float64 b, bool is_max, bool is_abs, int sign_ctrl,
         if (is_abs) {
             tmp_a = tmp_a & ~UINT64_C(0x8000000000000000); // clear the sign bit
             tmp_b = tmp_b & ~UINT64_C(0x8000000000000000);
-            signA = 0;
+            if (tmp_a != tmp_b)
+              signA = 0;
         }
 
         if (! is_max) {
```

If you confirm the patch is fixing the problem - I will merge it. Otherwise will have to wait until I write some unit test myself and try it.

bochs/cpu/softfloat3e/f32_range.cc
ergrelet: Hello,

tl;dr: yes the fix you proposed in your other message is working as expected, you can commit it and
close this PR if you want.

I'll answer your multiple messages at once here.

```
"For equal-magnitude operands with opposite signs, those tables require
MIN_ABS to select the negative operand and MAX_ABS to select the positive
operand."

So the issue is only about _ABS forms ?
```
```
If I understood the description correctly - the issue is ONLY when 2 arg are same magnitude and different sign.
Bochs doesn't obey the table from SDM:
```

Yes, both conditions need to be true. The thing is that an _ABS form *and*
equal-magnitude operands with opposite signs are required to trigger the bug.
More precisely, the bug is triggered when comparison-result sign control is used,
SRC1 is negative and SRC2 is positive. When SRC1 is positive and SRC2 is negative, the
original code already works properly. I think the confusion comes from the fact
that the fix proposed in the PR is misleading and cryptic I agree.  
The gist of it is to check:  
```
signA != signB && (!is_abs || equal_magnitude)
```
instead of:
```
signA != signB && !is_abs
```

The fix made it so Bochs shared the same branch/logic for both non-ABS and
ABS forms (vs non-ABS only before) for operands with opposite signs and of
equal magnitude as well as for non-ABS forms for operands with opposite sign
with operands of any magnitude (as before).  
But I agree, this code is not nice to read after the fix as the fix
shortcuts the logic by complexifying a seemingly unrelated condition instead of
doing a more readable shortcut in the next branch lower, like your proposed fix
does.

```
If so - try the following patch instead, it will be less intrusive (and therefore more correct) fix and also doesn't forget about float16:
```

From what I have tested, I confirm that your proposed patch fixes the bug
correctly, on top of fixing `f16_range` as well and making the overall fix
nicer to read for future readers:

> f16 MIN_ABS(-1,+1): old=3c00, patched=bc00
> f16 MAX_ABS(-1,+1): old=bc00, patched=3c00
> f16 MIN_ABS(-0,+0): old=0000, patched=8000
> f16 MAX_ABS(-0,+0): old=8000, patched=0000
>
> f32 MIN_ABS(-1,+1): old=3f800000, patched=bf800000
> f32 MAX_ABS(-1,+1): old=bf800000, patched=3f800000
> f64 MIN_ABS(-1,+1): old=3ff0000000000000, patched=bff0000000000000
> f64 MAX_ABS(-1,+1): old=bff0000000000000, patched=3ff0000000000000

Same way as I did for the initial validation section of this PR message. I also
used a harness that runs the actual VRANGEPS/PD/SS/SD instructions in a Bochs
guest to confirm the original and fixed behaviors (with your patch) end-to-end
as well. It covered ±1, ±0, both source orders, and immediates 4–7, checking
every packed lane and MXCSR flags.  
Here are input examples from the Bochs guest. These used comparison-result sign
control (`imm8[3:2] = 01`) and no masking.

| Instruction | imm8 | SRC1 | SRC2 | Unfixed Bochs | Expected / your patch |
|---|---:|---:|---:|---:|---:|
| `VRANGEPS` / `VRANGESS` MIN_ABS | `0x06` | `bf800000` (-1) | `3f800000` (+1) | `3f800000` (+1) | `bf800000` (-1) |
| `VRANGEPS` / `VRANGESS` MAX_ABS | `0x07` | `bf800000` (-1) | `3f800000` (+1) | `bf800000` (-1) | `3f800000` (+1) |
| `VRANGEPD` / `VRANGESD` MIN_ABS | `0x06` | `bff0000000000000` (-1) | `3ff0000000000000` (+1) | `3ff0000000000000` (+1) | `bff0000000000000` (-1) |
| `VRANGEPD` / `VRANGESD` MAX_ABS | `0x07` | `bff0000000000000` (-1) | `3ff0000000000000` (+1) | `bff0000000000000` (-1) | `3ff0000000000000` (+1) |
| `VRANGEPS` / `VRANGESS` MIN_ABS | `0x06` | `80000000` (-0) | `00000000` (+0) | `00000000` (+0) | `80000000` (-0) |
| `VRANGEPS` / `VRANGESS` MAX_ABS | `0x07` | `80000000` (-0) | `00000000` (+0) | `80000000` (-0) | `00000000` (+0) |

For comparison, the non-ABS MIN and MAX operations already produced the correct
result with exactly the same `-1,+1` operands:

| Instruction | imm8 | Unfixed Bochs | Expected |
|---|---:|---:|---:|
| `VRANGEPS` / `VRANGESS` MIN | `0x04` | `bf800000` (-1) | `bf800000` (-1) |
| `VRANGEPS` / `VRANGESS` MAX | `0x05` | `3f800000` (+1) | `3f800000` (+1) |

PS: If you were so irritated with the way my PR descriptions were written, you
could have mentioned it in the first PR, instead of waiting for the 14th one,
after 13 merged PRs. To me, it just looked like implicit agreement that
the PRs were fine as is. But now that you made it clear, I'll surely refrain
from sending such PRs in the future.

stlintel:
Just to avoid any misunderstanding.
The things you doing are highly valuable and very appreciated. Whatever tool you use to find these issues is great and please continue doing do.

I only had issues with the description written by AI. In many cases it was very hard to understand and only the code fix itself was clear enough to tell what's going on. The ideal description should be short, showing the test case (instruction and inputs) which is misbehaving and why. Ideally with reference to manual. Instead of everything written above tell:

Xmm0 = 0.0
Xmm1 = -0.0
Vrangess xmm0, xmm1

Expected: x
Got: y

This would be more than enough.

Thanks for your work anyway.

stlintel:
Fixed by 04b100ae1f955d531de418ee9fd1332c1a6d7b86

PR #865, opened 2026-09-11
implement HOST SIMD intrinsics optimization

stlintel:
Let Bochs SIMD instruction handlers to use HOST intrinsics when available - can significantly speedup simulation of heavy SIMD workloads

I will do more performance testing and ask you all to see if the idea working not only with gcc and Windows.
What happens in Linux ? MSVCPP ? Clang ? Old hardware ?


stlintel:
Done with Claude Code. 

<html>
<body>
<h3 dir="ltr">Speedup: measured, not guessed</h3><p dir="ltr">I benchmarked <code>ref::</code> (plain C loop) vs <code>intrin::</code> (current <code>simd_int.h</code>) side by side, same function, same inputs, 200M iterations each, with each iteration's operand perturbed by the loop counter (needed — my first attempt let GCC prove the loop was idempotent/invariant and hoist it away entirely, which is how I got a bogus "16-million-x" reading for <code>pshufb</code> the first time).</p><p dir="ltr"><strong>Default build flags — what actually ships today</strong> (SSE2 tier only):</p><div dir="ltr"><div>
function | plain C | intrinsic | speedup
-- | -- | -- | --

- xmm_paddb | 4.81 ns | 4.03 ns | 1.19x
- xmm_paddsb (saturating) | 8.72 ns | 4.26 ns | 2.05x
- xmm_pminub | 4.66 ns | 4.30 ns | 1.08x
- xmm_packuswb | 6.80 ns | 4.03 ns | 1.69x
- xmm_pmaddwd | 9.18 ns | 4.66 ns | 1.97x
- xmm_psadbw | 6.91 ns | 4.06 ns | 1.70x
- xmm_psraw (variable shift) | 3.16 ns | 0.40 ns | 7.88x

</div></div><p dir="ltr"><strong>With <code>-march=native</code></strong> (unlocks SSSE3/SSE4.1/AVX2 on this CPU): <code>pshufb</code> jumps to <strong>38x</strong> (SSSE3's single-instruction table shuffle vs. 16 branchy scalar lookups) and <code>psraw</code> to 9.35x — but <code>phaddw</code> came out <strong>slower</strong> (0.52x) here, which is a real, worth-noting anomaly: <code>PHADDW</code> has nontrivial latency on real hardware and this micro-benchmark serializes each call on the previous result, so it's measuring latency, not throughput; a real workload issuing independent <code>phaddw</code>s back-to-back wouldn't see that penalty. I'm not papering over it — it's a legitimate case where "one instruction" isn't automatically "faster."</p><h3 dir="ltr">The honest caveat</h3><p dir="ltr">These are <strong>per-call, everything-hot-in-cache</strong> numbers for the SIMD helper in isolation. In the actual Bochs interpreter, one call to <code>xmm_paddb</code> sits behind instruction fetch, decode, dispatch, operand fetch from the register file, and exception/flag bookkeeping — the SIMD helper itself is a small slice of the total per-instruction cost. So a 2x speedup on the helper does <strong>not</strong> translate to 2x faster overall emulation of SSE-heavy code — realistically more like low single-digit <strong>percent</strong> improvement on SSE/MMX-heavy workloads, and effectively nothing on workloads that aren't SIMD-heavy. The functions that show the biggest wins here (saturating arithmetic, <code>pmaddwd</code>, variable shifts, <code>pshufb</code>) are the ones where the scalar reference does real per-element branching/saturation logic — that's exactly where a vector instruction earns its keep; the ones we already reverted (<code>andps</code>, <code>unpck*</code>) confirm the opposite case correctly at ~1.0x.
</body>
</html>

Vort:
1. Right now Bochs conveniently has 1 binary per platform. How is it going to change with such PR merged? There are several possibilities (Windows examples):
1.1. Binary is tuned to use SSE2 intrinsics and can be used as usual. However more advanced CPU capabilities are left unutilized;
1.2. Binary is tuned to the most popular set of intrinsics available. More performance, but some users won't be able to use binary at all;
1.3. Bochs is distributed with several binaries, most appropriate one is chosen during installation. Most performance, all users can use Bochs, but installer size will increase and build/packaging process will become more complicated;
2. Merge of such platform-specific code will mean that some users will have one branch of code, other users will have other branch of code activated. It means problem in searching of bugs, because there is no common execution path anymore;
3. Alternatives:
3.1. Did you considered runtime detection of CPU features? Will it just make emulation unconditionally slower? This is how OpenSSL works for example: a single binary is used, yet processor-specific optimizations are applied when possible;
3.2. The most promising approach in my opinion is figuring out how automatic vectorization in compilers works. It may be possible to change code a little somehow and compiler will figure out by itself which corresponding SIMD instruction should be used. But that's a theory, I don't know to which extent it is possible in reality.

stlintel:
>The most promising approach in my opinion is figuring out how automatic vectorization in compilers works. It may be possible to change code a little somehow and compiler will figure out by itself which corresponding SIMD instruction should be used. But that's a theory, I don't know to which extent it is possible in reality.

Yes, I agree this is the most promising direction. But so far gcc auto-vectorizer was pretty dumb and was able to pick up almost nothing, only most trivial and idiotic cases.
I am still using gcc 14 compiling with -O3.

gcc 15 and 16 claim some extra improvements for auto-vectorizer but for some reason I am still sceptical.

Regarding many binary distributions.
I don't think it is important.

At the end it is expected to be a small speedup (execution itself is probably only 30%-50% of overhead there). 64-bit binary already has SSE/SSE2/SSE3/SSE4.1/SSE4.2 by definition and nearly everyone already has AVX. I think today AVX/AVX2 is the industry standard. We can compile up to AVX/AVX2 only. 

Other users which do not have AVX in their host (older than Haswell let's say - June 4, 2013) will have to get source code and compile by themselves.

stlintel:
Determining at runtime is possible but and even cheap. 
On startup determine host capabilities and decide which CPU optimizations to enable.

But this means we need to have 2 versions of each such function handler - with intrinsic and without - and they are selected during construction of CPU decoding tables once.
But currently I am struggle with too many CPU functions already. 
If I will be able to find a way how to manage it relatively easily (like templates again?) - may be.

Vort:
> But so far gcc auto-vectorizer was pretty dumb and was able to pick up almost nothing, only most trivial and idiotic cases.

I was able to find interesting case. Here is how `BX_CPU_C::HANDLE_SSE_2OP<xmm_punpckhwd>` looks like with `bochs-fdaa285.exe.zip` (no intrinsics).
I wonder why similar `xmm_punpcklwd` function has no SIMD instructions. Will uncommenting of this line make code actually faster?
https://github.com/bochs-emu/Bochs/blob/39f65c517cd228aff069a4c913639948bd94c69a/bochs/cpu/simd_int.h#L250

изображение

PR #866, opened 2026-09-12
Implement texture filtering for GeForce

Vort:
This change allows programs to use linear interpolation for texture filtering.
3DMark 99 Max shows smooth checkerboard edges with 81.98 driver and NV40.
Windows 7 correctly blurs wallpaper behind taskbar with 309.08 driver and NV40.
bochs_gf40_3dmark99_trilinear
bochs_gf40_windows7_linear

### DIFF
--- a/bochs/iodev/display/geforce.cc
+++ b/bochs/iodev/display/geforce.cc
@@ -3193,6 +3193,7 @@ void bx_geforce_c::d3d_texture_process_format(gf_texture* tex)
     case 0x00: // Y8
     case 0x01: // AY8
     case 0x0b: // I8_A8R8G8B8
+    case 0x19: // A8
     case 0x1b: // AY8
     case 0x81: // B8
       tex->color_bytes = 1;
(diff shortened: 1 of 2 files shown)


stlintel (approved):

sskras:
@Vort, you must be doing a lot math during the reverse engineering, I guess, be it only reading the specs and some source code, or just reversing some closed code. That's incredible, IMO. Hats off to you :)

Vort:
Surprisingly, most of the information this time I got from OpenGL 2.0 spec.
Imagining 4-dimensional interpolation wasn't easy however.

Fun fact: this change required me to write test program ([TexFilter.zip](https://github.com/user-attachments/files/32148953/TexFilter.zip)) and it revealed that old GeForce cards have one feature which my newer HD Graphics 4600 does not have. Looks like OpenGL 3.0+ cards dropped support for `GL_CLAMP` wrapping mode, so I wasn't able to look how test program should behave without booting computer with real GeForce card.

PR #875, opened 2026-09-18
Add Apple SMC emulation and SMBIOS table generator for macOS guest support

stlintel:
Bochs currently has no way to run x86 macOS as a guest: VirtualBox and VMware both required (and still require) specific hardware emulation that a real Mac provides and generic PC firmware does not. This is a first step toward closing that gap, targeting the OVMF/fw_cfg boot path (El Capitan - Mojave era, Haswell-class CPU profile such as iMac15,1).

Two pieces of Apple-specific hardware are added:

- Apple SMC (System Management Controller) emulation (iodev/applesmc.cc/h), ported from QEMU's hw/misc/applesmc.c. This is I/O-port hardware (0x300/0x304/0x31E) that AppleSMC.kext talks to during boot; macOS refuses to proceed past early boot without a responding SMC. The OSK0/OSK1 proprietary Apple key is never hardcoded - it defaults to a placeholder and is configurable via a new bochsrc "applesmc: osk=..." option.

- An SMBIOS table generator (iodev/smbios_tables.cc/h) producing types 0/1/2/3/127 (BIOS, system, baseboard, chassis, end-of-table) with an Apple-identifying profile (manufacturer "Apple Inc.", product "iMac15,1", board id "Mac-42FD25EABCABB274"). macOS's installer and kernel read these to decide whether they're running on Apple hardware. Delivered to OVMF via fw_cfg's existing etc/smbios/smbios-tables and etc/smbios/smbios-anchor files, the same mechanism QEMU uses, so no new loader/checksum machinery was needed.

- bios/acpi-dsdt.dsl / iodev/acpi_dsdt_aml.h: add a Device(SMC) node (_HID "APP0001", I/O range 0x300-0x301, IRQ 6) so the SMC device is discoverable via ACPI, matching QEMU's isa-applesmc AML shape so AppleSMC.kext's existing ACPI matching keeps working unmodified.

Build system: new --enable-applesmc configure flag (default on), BX_SUPPORT_APPLESMC wiring through config.h.in/configure.ac, plugin registration (plugin.cc/h, iodev/iodev.h, iodev/devices.cc), and Makefile.in updates for both the SMC plugin and the (unconditional, like acpi_tables.o) smbios_tables.o.

Status: this is NOT a complete, verified macOS boot path yet. Real gaps remain:
- CPU identity: needs to be set to a period-accurate profile (e.g. corei7_haswell_4770) per guest target.
- Storage: no AHCI/SATA in Bochs, so the installer needs to be exposed through USB Mass Storage or similar instead - not yet tested.
- EFI NVRAM persistence: Bochs loads OVMF as a single flat ROM with no write-back path, so boot-args/csr-active-config writes don't survive a reboot. Not required for a first boot attempt but will matter for real installs, which restart partway through.
- No full install has been completed against real El Capitan/Mojave media yet

Posting this now, ahead of a full working boot, so the SMC/SMBIOS groundwork can get wider testing and review. Feedback and testing against real macOS installer media welcome.


stlintel:
Folks as non-Mac person, I had trouble to convert DMG file to smth I can test and boot.
I will appreciate if someone can test it too.

Randrianasulu:
There is dmg2img utility:

https://github.com/lekensteyn/dmg2img

or may be you can use qemu-img convert for converting install dmg to some boot partition and EFI partition.

I used 10.6.0 Intel dmg for qemu-based hackintosh, but that one was installing on q35-2.11 machine (SATA)

Randrianasulu:
I think 10.5.4 installer should be bootable on IDE CD ?

https://github.com/adespoton/utmconfigs/issues/71
https://github.com/LongQT-sea/OpenCore-ISO/releases
(I used version 0.2 iso)

stlintel:
> I think 10.5.4 installer should be bootable on IDE CD ?
> 
> [adespoton/utmconfigs#71](https://github.com/adespoton/utmconfigs/issues/71) https://github.com/LongQT-sea/OpenCore-ISO/releases (I used version 0.2 iso)

It is bootable CD, but how it install anything ?
This is what it knows to do:

image

And 15M ISO doesn't look like installer of anything

Randrianasulu:
> > I think 10.5.4 installer should be bootable on IDE CD ?
> > [adespoton/utmconfigs#71](https://github.com/adespoton/utmconfigs/issues/71) https://github.com/LongQT-sea/OpenCore-ISO/releases (I used version 0.2 iso)
> 
> It is bootable CD, but how it install anything ? This is what it knows to do:
> 
> image
> And 15M ISO doesn't look like installer of anything

I added second cdrom with installer (in qemu) and it showed up there ... but if you can change cdrom on the fly in Bochs you can try this method?)

stlintel:
> > > I think 10.5.4 installer should be bootable on IDE CD ?
> > > [adespoton/utmconfigs#71](https://github.com/adespoton/utmconfigs/issues/71) https://github.com/LongQT-sea/OpenCore-ISO/releases (I used version 0.2 iso)
> > 
> > 
> > It is bootable CD, but how it install anything ? This is what it knows to do:
> > image
> > And 15M ISO doesn't look like installer of anything
> 
> I added second cdrom with installer (in qemu) and it showed up there ... but if you can change cdrom on the fly in Bochs you can try this method?)

I am going to family vacation till end of this month so hope someone more experienced can test and provide exact instructions where and if it stuck + detailed instructions to reproduce.

stlintel:
This one was really crazy :

In leaf 4: corei7_haswell_4770.cc:359 returns EAX = 0x1C004121 for every cache level. Bits 31:26 of that value are 000111 = 7. Intel defines that field as "max addressable core IDs in the package − 1", so it means 8 IDs. On the real i7-4770 those 8 IDs are 4 cores × 2 hyperthreads, not 8 physical cores.

In macOS: the kernel's cpuid_set_info at ffffff80003bc58b shifts EAX right by 26, adds 1 and stores the result as core_count. So macOS reads 8 there, whatever the guest actually has.

Thread count: it comes from the other side. cpuid_set_info sets thread_count to 1 unless the CPU's HTT flag (CPUID.1 EDX bit 28) is set. Only then does it use EBX[23:16] from leaf 1. Bochs's leaf 1 EDX is 0xAFEBFBFF, so HTT is off, and thread_count = 1.

So the numbers macOS sees are core_count = 8, thread_count = 1, and the kernel divides thread_count / core_count, which gives 0. That confirms the diagnosis. Two corrections to what I told you earlier:

Leaf 1's logical-processor count isn't the source of the 1; the missing HTT flag is.
My fix still works: with one configured CPU, leaf 4's core field becomes 0, so core_count is 1 and the ratio is 1/1

Randrianasulu:
qemu code now grow into hard to follow codebase? ;)

on Leopard install disks there is #36 iso, 7Gb

https://macintoshgarden.org/apps/mac-osx-mac-os-10-ppc

